PR #5, opened 2022-04-26
adds placeholder view to force rendering regions

### LINKED ISSUE #4: Temporarily add placeholder view/blocks to force rendering regions
If there are no blocks in a region, the region doens't render. This adds a placeholder view to render basically nothing except the region itself, so we can work on theming it.

markconroy:
Fixes #4


markconroy:
Not needed any more.

PR #10, opened 2022-05-05
Enable Page content as Group plugin and remove Subites

### LINKED ISSUE #9: Add page content type as group plugin and remove subites

stephen-cox:
Fixes #9


config/optional/core.entity_view_display.group_content.group_content_type_1a9b2b47fec9e.default.yml
```diff
@@ -0,0 +1,14 @@
+uuid: 97803c07-9521-443e-90a7-786ae3ff1cd9
```
markconroy: I think we should remove the UUID from these files.

config/optional/core.entity_view_display.group_content.group_content_type_1a9b2b47fec9e.default.yml
```diff
@@ -0,0 +1,14 @@
...
+langcode: en
+status: true
+dependencies:
+  config:
+    - group.content_type.group_content_type_1a9b2b47fec9e
```
markconroy, line 5: I'm not sure we can do anyting here, but I'd much prefer if these said `group.content_type.group_content_type_node_page` instead of the hash at the end.

localgov_microsites_group.info.yml
```diff
@@ -4,10 +4,11 @@ description: Group integration for Microsites.
...
 dependencies:
+  - field_group:field_group
   - group:group
+  - group:gnode
   - image:image
   - link:link
   - media_library:media_library
   - path:path
-  - field_group:field_group
   - localgov_core:localgov_media
```
markconroy, line 14: Do we need to add `localgovdrupal/localgovpage` to this module's composer.json?

stephen-cox: The config for this is in optional, and so we don't need to add an explicit dependency, the config will only get installed if localgov_page is enabled (which is done in the profile). This keeps the dependency chain simpler and shouldn't cause issues, but we could make the dependency explicit.

config/optional/core.entity_view_display.group_content.group_content_type_1a9b2b47fec9e.default.yml
stephen-cox: I've tried renaming these, but then I hit a strange Field UI error, so it doesn't look as easy as just renaming the IDs and the files.

localgov_microsites_group.info.yml
ekes: Correlated with https://github.com/localgovdrupal/localgov_microsites_group/issues/13 that I just posted.

Is the assumption going to be that we always have localgovpage; and it will _always_ be enabled?
Or does it make sense to accept that people might not use it and have directory only groups or some such?

ekes (approved):
Other than https://github.com/localgovdrupal/localgov_microsites_group/pull/10#discussion_r866609858 which we can maybe push into the other issue I mentioned. I'd say this is good to merge. I've just run it against a fresh install as well and all working as expected.

PR #16, opened 2022-05-06
Add main and utility group menus

stephen-cox:
Part of localgovdrupal/localgov_microsites#39

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,6 +7,7 @@
     "drupal/core": "^9",
     "drupal/field_formatter_class": "^1.5",
     "drupal/group": "^1.4",
+    "drupal/group_content_menu": "^1.1",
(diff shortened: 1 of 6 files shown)


markconroy (approved):

ekes:
I was going to suggest adding the default admin permissions to this PR. But I'm going to spin it out into a new issue #26 

At the moment the menu content menu only exposes permissions for the menu itself (confusingly identified as for _one_ menu when it's all the menus) and doesn't expose any permission for adding menu items (and uses the rather far-reaching `Administer group` permission for this).

So approving as all the rest works, and we need to sort permissions elsewhere.

ekes (approved):
see #26 for follow-up.

Issue #18, opened 2022-05-06
Add PHPunit (functional) tests for microsites creation wizard


ekes:
Need to step through the form every which way - with different group type settings for membership.

willguv:
@ekes @finnlewis has some of this been covered already?

finnlewis:
See https://github.com/localgovdrupal/localgov_microsites_group/pull/226 and https://github.com/localgovdrupal/localgov_microsites_group/issues/225

Issue #15, opened 2022-05-06
Create Microsites Wizard: Configurable domain group settings.


ekes:
[For sprint 5+]

Which plugin, and fields are show on the domain settings are hard coded, could be made configurable.

As part of the wizard form, we want to be able to offer some further configuration options to allow the microsites controller to decide: 
 - which content types are available
 - what permissions are being deployed

ekes:
Related #20

willguv:
@ekes @finnlewis are we doing a wizard now? Feels like this card can go

Issue #14, opened 2022-05-06
Create Microsites Wizard: Select which content types are enabled


ekes:
Blocked on having more content types for now.

ekes:
We have the content types. Suggested path forward:
 * All content types are enabled on all grousp
 * To add permissions to access content types per group
 * Make a nice for to do this.
Follow up from https://github.com/localgovdrupal/localgov_microsites_group/pull/138 and https://github.com/localgovdrupal/localgov_microsites_group/pull/137 to appropriately remove `/events` and `/news/search` depending if the content type is deemed enabled.

Issue #13, opened 2022-05-06
Create Microsite Wizard: Create first site page


ekes:
Does this mean we _always_ enable the page content type?

Down the line should what initial page is created (like a directory as alternative) be configurable.

PR #19, opened 2022-05-06
Feature/1.x/microsite creation wizard

ekes:
This seems to be at a stage it would be good to get into the main branch:-

Adds a /admin/microsites page, and menu item
Adds /admin/microsites/add/microsite with wizard to create the group. Displayed as an action link on the /admin/microsites page.

The group form in the wizard (and optionally, but default disabled, membership form) are configured for fields in the 'new_domain' display mode. Included in the install configuration.

The group form exposes the owner (entity reference autocomplete form) this is used for the first admin member of the group, and the content auto generated in it.

The fields for the group domain site configuration are coded into the wizard output for now.

Follow-ups:-

 #11 Make it work with more than one group type - probably lower priority, but not too time consuming if done _now_ rather than in the future, might work best with a change to domain group 3.x too
 #14 Add a form for setting with content types are enabled for the group initially. Completely independent of other work. No order dependency except...
 #13 which is done in this PR assumes the localgov page content type
 #12 will make the /admin/microsites page more useful
 #15 is adding the configuration to the domain group configuration hard coded as mentioned above
and
 #18 if the wizard is doing much we want given above should get it tested with all variant.


config/optional/views.view.localgov_microsites_overview.yml
```diff
@@ -0,0 +1,253 @@
+uuid: ba5220bd-c461-42f9-b23b-745644c47175
```
stephen-cox: Rogue UUID to remove

config/optional/views.view.localgov_microsites_overview.yml
```diff
@@ -0,0 +1,253 @@
...
+langcode: en
+status: true
+dependencies:
+  config:
+    - group.type.microsite
+    - system.menu.admin
+  module:
+    - group
+_core:
+  default_config_hash: iwN3wTjbnaYoCsjWYqpSSur_nI1glzMDTaSh2XkDdwE
```
stephen-cox: Also need removing

stephen-cox (approved):
The basic concept looks sound. What's on the steps will need looking at.

We'll also need to set some basic permissions to allow access to group content, but this could be done as part of simplifying the permissions system. If this isn't going to happen next sprint perhaps we need a published checkbox, or use the Domain status, to set some basic defaults.

Happy for this to be merged now and add to it. Just found a couple of rogue config UUIDs to remove first.

Issue #35, opened 2022-05-16
Add menu bar for microsite admins to be able to find their group


finnlewis:
When a 'microsite admin' logs in we want to redirect them to the group page.

Like /group/1

However, if they are elsewhere, we need to offer a link or links so that they can manage their content and config (admin toolbar ? custom block displaying the tabs of the group?)

ekes:
@markconroy @msayoung Before I make a blank-ish template for this does the render array make sense to you https://github.com/localgovdrupal/localgov_microsites_group/commit/5b881e96ac64218ff3e1773505964400ebc13393#diff-31a6feeb998ebe17cbedc629279f120b16a3db15468702d0ca85b21e486654ddR118=

It's basically an array of links with #access and #weight.

markconroy:
@ekes  

1. Do we have a fall back for if an admin logs in and is an admin of more than one group? I'd expect most content designers in a council will be admins on many of the groups/microsites
2. The render array you are proposing looks like a good idea to me.

ekes:
So far the understanding is that a user who is to edit content logs into the microsite url. So there is only one group, the one for that microsite hostname. The menu displayed are the local tasks for the group based on the hostname. If they want to edit another one they need to log into that microsite. 

On the control site it is possible to enter content, from within the group context, but it's been built round the controller so far.

markconroy:
That makes sense, but I'd presume since this is for councils that most editors will be editors on more than one site and would expect to login just once to their microsites portal. 

If the original thought was that this was going to be part of LGD rather than a separate instance, they'd be managing all their content from within their LGD site

finnlewis:
Interesting assumption @markconroy - I had assumed the opposite, that they would want to log into the site they are managing to keep it all in context. 

We should add these competing assumptions to the assumptions log and ultimately confirm if it is one or the other, or both! 

Meanwhile, does this assumption have a material impact on how the localised group menu works @ekes ?

finnlewis:
Either way, we will need to style the block.

If I place the 'Microsite tasks' block in the 'tabs' region, then log into Microsite 1 as admin1, it looks like this: 

(you can just see the links against the dark background)

![Image](https://user-images.githubusercontent.com/326588/169474327-eaba5458-902a-4e4b-822f-8aefdf509872.png)

@msayoung one for you?

ekes:
> Meanwhile, does this assumption have a material impact on how the localised group menu works @ekes ?

[...]

PR #34, opened 2022-05-17
Feature/1.x/25 login redirects

ekes:
Adds redirect for login based on control or microsite domain.
Also adds tests @stephen-cox guessing they might need some additional domain magic - or is it all there?

PR #40, opened 2022-05-18
Microsite admin menu based on local tabs for group

ekes:
From #35 

It does not place the block, just creates it.


ekes:
Ach. and includes #34 might as well get that merged.

finnlewis:
Looking at this now

finnlewis (approved):
Yeah man, works for me! 

We'll need to decide where and how to place this block, maybe on https://github.com/orgs/localgovdrupal/projects/21/views/2?filterQuery=block @stephen-cox 

But might as well get this merged.

PR #39, opened 2022-05-18
Feature/1.x/79 help text

msayoung:

Issue #38, opened 2022-05-18
Remove /admin/microsites from microsite


ekes:
- [ ] Remove microsites admin list from microsite
- [x] ~~Error on microsites admin if accessed from a microsite.~~
```
Drupal\Core\Database\DatabaseExceptionWrapper: Exception in Microsites[localgov_microsites_overview]: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'base_table.id' in 'where clause': SELECT "groups_field_data"."id" AS "id" FROM "groups_field_data" "groups_field_data" WHERE (("groups_field_data"."status" = :db_condition_placeholder_0) AND ("groups_field_data"."type" IN (:db_condition_placeholder_1))) AND ("base_table"."id" = :db_condition_placeholder_2) LIMIT 51 OFFSET 0; Array ( [:db_condition_placeholder_0] => 1 [:db_condition_placeholder_1] => microsite [:db_condition_placeholder_2] => 1 ) in main() (line 19 of index.php). 
```

ekes:
The other is hopefully fixed, without breaking anything else (tests still passing) on domain_group
https://git.drupalcode.org/sandbox/ekes-3278349/-/commit/82875634727eb11f404c616932bbb0d7e080f5aa

That you shouldn't be able to access it on a microsite is surprisingly more complicated. The view access on domain requires knowing the domain hostname; there isn't an access filter for 'default' 'not default' domain. So I guess to remove this we'll have to write one.

Updating the issue title for this issue.

finnlewis:
I think this fixed @ekes , no?

Issue #23, opened 2022-05-06
GroupContent as the site content; or Node 


ekes:
The GroupContent (a content relation between the group and its content - in this case Node) at the moment is just that, and displays only the title. The content that is used for the site is the node itself. This is also set for the permissions (note #20) so the node is visible to the anonymous user.

Discussion:-

It is possible to load and display the Node on the GroupContent and make this the 'content', hiding the original Node. This was something that @markconroy mentioned they are doing with some of the University sites. It means the content /node/_x_ is also visible on /group/1/content/_y_ and even on /group/2/content/_z_. In a single site you then want to rabbit hole the /node/_x_ with our set up this wouldn't even be needed.
In our use case only the GroupContent /group/1/content/_y_ is visible on the domain assigned to group 1; and it would even be possible to have the same node on GroupContent /group/2/content/_z_ on the domain assigned to group 2. Path aliases would also be set on the GroupContent per group.

@markconroy can you remember: were there any adverse UX issues for editors, or other quirks; issues for theming; loading content, or fixes that you needed to make this a good option?

Generally it seems like it might be a good idea.

markconroy:
@ekes to get more than just the title to display, you need to go to 
Groups > Microsites > Content > Page > Manage Display > Title 

And set "Title" as "rendered entity", and choose view mode "Full".

There wasn't any major issue with theming. One issue we had was that since we (the university site I was working on) have our page titles inside our full view mode templates for semantics, the page title was getting rendered 2 times - via block AND via our template. But that's an easy fix. 

The other thing was that there were 2 article tags, one from the group-content.html.twig file and one from the node.html.twig file, so we ended up with this:

```
<article class="group-content">
  <article class="node">
  </article>
</article>
```

To get aroudn that, I just removed everything from the group-content template and let that render only `{{ content }}`

ekes:
One more vote for Group Content https://github.com/localgovdrupal/localgov_microsites_group/pull/41 as it saves having to do anything complicated to add the group context for the menu if everything is displayed under the group context anyway.

ekes:
Questions: Aliases, paths, breadcrumb, admin theme

When in the group context the breadcrumb becomes something like groups > group name > content
However we don't want to include the group in the breadcrumb if it's shown on a domain. The piece of content could be the homepage and then other content could be under it.

[...]

PR #41, opened 2022-05-19
Add block config for main and utility menus

stephen-cox:
Regions are added to the theme with this PR: https://github.com/localgovdrupal/localgov_microsites_base/pull/10

Closes  #39


stephen-cox:
Group menus aren't rendering. This looks to be an issue with the theme working with Groups. Looking into it now

ekes:
This makes an (optional) but circular dependency on localgov_microsites_base theme?

The theme has to depend on this module, the module doesn't have to depend on the theme. So you only get this optional configuration if the theme is there, this module will always be there if the theme is there. So it would make sense for this to be in the theme?

finnlewis:
> Group menus aren't rendering. This looks to be an issue with the theme working with Groups. Looking into it now

Yeah, I'm not seeing menus after adding links to them on the microsite: https://localgov-micro-1.lndo.site/ 

But the blocks do get placed on a fresh install, so that's nice :)

stephen-cox:
I was a bit hasty creating this PR. There's a fundamental issue placing the Group Content Menu blocks.

To determine the group it use the GroupRouteContext which comes with Group the module. This determines the group from the current route and doesn't work with domains.

The easiest fix looks to be to create a custom Block plugin that inherits from the Group Content Menu block and sets the group rather than using the context. It would be nice if we could just override the context but this doesn't seem possible.

Inheriting the block and modifying behaviour seems a little flaky, but works for now. Other thoughts on solutions to this are welcome.

ekes:
Options: 
- adding more contexts; we certainly need them domains; for example there isn't one for the default domain :-/ (at least not without knowing what the hostname is already - which isn't good for a generic view) #38
- only show content in the group context, so for example content would have to be displayed on the group content relationship #23

stephen-cox:
Thanks @ekes, I have updated the code here to make it clear it needs replacing or rewriting once we have agreement on how to determine content context. I've also moved the block placement config to the theme with https://github.com/localgovdrupal/localgov_microsites_base/pull/11

ekes (approved):
Happy with this as a temporary fix.

Think answering the questions on #23 might be the best solution when we start again after testing.

finnlewis:
You happy to merge this @stephen-cox or do you want me to test it too?

stephen-cox:
@finnlewis Merged it - you can test it with this PR https://github.com/localgovdrupal/localgov_microsites_base/pull/11

PR #49, opened 2022-05-20
Patch to group module to check create in addtion to add permissions for node to group.

### LINKED ISSUE #45: Admins don't get create content form linked, while they have access to it.
That's a bug. They do have the permission to add a page, just don't get to that form, or at least on my site. So new issue I'd say, looking at it:

If you go to https://localgov-micro-1.lndo.site/group/1/content/create/group_node%3Alocalgov_page
(and are the admin of it you can add a page).

The action button on https://localgov-micro-1.lndo.site/group/1/nodes is visible too

But it goes to https://localgov-micro-1.lndo.site/group/1/node/create which would list the content types, which it doesn't, and also as there is only one content type it should automatically redirect to https://localgov-micro-1.lndo.site/group/1/content/create/group_node%3Alocalgov_page which it also doesn't.

_Originally posted by @ekes in https://github.com/localgovdrupal/localgov_microsites_group/issues/26#issuecomment-1131854737_

ekes:
It's a patch so requires composer install.

Fixes https://github.com/localgovdrupal/localgov_microsites_group/issues/45

But we should write tests for it for group module and post to
https://www.drupal.org/project/group/issues/2842630 to get it committed.
Other issues on #45 can be referenced as I'm sure they're all dupes.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,6 +19,9 @@
     "patches": {
       "drupal/group_content_menu": {
         "Menu item create permisson #3280653": "https://www.drupal.org/files/issues/2022-05-16/3280653-01-menu-item-add-permission.patch"
+      },
+      "drupal/group": {
+        "Empty page when trying to create group node localgovdrupal/localgov_microsites_group#45": "https://www.drupal.org/files/issues/2019-01-04/2842630-20.patch"
       }
     }
   }


finnlewis (approved):
Let's merge it!

PR #54, opened 2022-05-20
Retemplate to allow the .tabs class to work

msayoung:


ekes (approved):
Anything is better than my quick default :)

Issue #55, opened 2022-05-20
New permissions for controller


ekes:
Do we want to add some permissions for the controller. For example viewing the 'overview'. Which is presently looking for the override group permissions.

PR #56, opened 2022-05-20
Restrict access to microsites overview.

ekes:
Add a permission to restrict access to the microsites overview. #55 as follow-up if it's better to add some custom permissions.


finnlewis (approved):
Yup - this works.

PR #73, opened 2022-05-23
Inital fix to replicate exisitng content into a group as created.

ekes:
Notes:-

Uses config localgov_microsites_group.settings[default_group_node]
which is not set, nor is there a UI to set it, defaults to node 1.
Follow-up: Add UI, default value in profile to the imported default
content.

Replication of entity with paragraphs will always cause them to be
cloned, no other referenced content will be cloned. Questinon (and todo)
in the code to decide if this should be restricted (by some value on the
entity).


finnlewis:
So far I keep getting: 

The service "localgov_microsites_group.default_content" has a dependency on a non-existent service "replicate.replicator".

ekes:
Ach.. module not enabled I guess.

ekes:
There we go. Nicely rushed out and missed.

finnlewis:
Having trouble testing this.

In my root composer.json I've got: 
```
        "localgovdrupal/localgov_microsites": "dev-feature/replicate as 1.x-dev"
```

Which should get https://github.com/localgovdrupal/localgov_microsites/tree/feature/replicate and then chain on to 

```
"localgovdrupal/localgov_microsites_group": "dev-feature/1.x/replicate-default-content-new-group as 1.x-dev",
```

But I'm still stuck with dependency fun:

```
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - localgovdrupal/localgov_microsites_base 1.x-dev requires localgovdrupal/localgov_microsites_group 1.x-dev -> satisfiable by localgovdrupal/localgov_microsites_group[1.x-dev].
    - You can only install one version of a package, so only one of these can be installed: localgovdrupal/localgov_microsites_group[dev-feature/1.x/replicate-default-content-new-group, 1.x-dev].
    - localgovdrupal/localgov_microsites dev-feature/replicate requires localgovdrupal/localgov_microsites_base 1.x-dev -> satisfiable by localgovdrupal/localgov_microsites_base[1.x-dev].
    - localgovdrupal/localgov_microsites dev-feature/replicate requires localgovdrupal/localgov_microsites_group dev-feature/1.x/replicate-default-content-new-group as 1.x-dev -> satisfiable by localgovdrupal/localgov_microsites_group[dev-feature/1.x/replicate-default-content-new-group].
    - Root composer.json requires localgovdrupal/localgov_microsites dev-feature/replicate as 1.x-dev -> satisfiable by localgovdrupal/localgov_microsites[dev-feature/replicate].

Use the option --with-all-dependencies (-W) to allow upgrades, downgrades and removals for packages currently locked to specific versions.

ekes:
I read that as localgov_microsites has the requirement  `"localgovdrupal/localgov_microsites_group": "dev-feature/1.x/replicate-default-content-new-group as 1.x-dev"` It should also just depend on `1.x-dev` with the alias in the root.

finnlewis:
I'm a bit stuck with this

finnlewis (approved):
Beautiful! Works for me, let's merge it!

ekes:
Now also tested with new profile default content and merged with 1.x as of now. Sorted.

PR #78, opened 2022-05-23
Fixes #77 for new installs.

### LINKED ISSUE #77: Alias field on GroupContent User (member) form
Hi @ekes 

On this one, could we remove or hide the 'URL alias' from the user add form?  I don't think we need it and it is bound to cause confusion.

Image

_Originally posted by @finnlewis in https://github.com/localgovdrupal/localgov_microsites/issues/37#issuecomment-1134551989_

ekes:


finnlewis:
Hi @ekes 
I still see the url alias field on a fresh install.
What have I missed?

ekes:
The randomness of which configuration get installed. Guess because it's only created with the microsite group type it's not going to install the config, so we need to do this in code :-(

ekes:
Actually moving the config to /install/ rather than /optional/ with its dependencies (and removing the one that group actually already provides) fixes it for installation for me.

finnlewis (approved):
Yay! It werks!

finnlewis:
Merged! Thanks @ekes

Issue #74, opened 2022-05-23
Microsites controller: create a new site [remaining issues]


ekes:
Hi @ekes 

On this one, a couple of things to do: 

- [ ] Rename 'Edit' tab to 'Site design'
- [ ] Hide or remove the 'Revision Information' tab from the edit form

![image](https://user-images.githubusercontent.com/326588/169810272-127c2ec6-f73e-490f-b76d-f56a3f2bc65f.png)

_Originally posted by @finnlewis in https://github.com/localgovdrupal/localgov_microsites/issues/36#issuecomment-1134558416_

ekes:
> Rename 'Edit' tab to 'Site design'

So that's fixed in bd4fd8c I've just pushed it, a c&p error in the refactor.

> Hide or remove the 'Revision Information' tab from the edit form

Seems Group doesn't offer permissions for this. Might it be best to just set it default to not make a new revision. Then the slider is still there in grey, but nothing else. Anything less is going to be a form alter that I fear will cause pain to someone working out where it's done when we do need revisions.

ekes:
@finnlewis should have tagged you on that question :point_up:

PR #79, opened 2022-05-23
Disable default revisions instead of removing #74

ekes:
From suggestion https://github.com/localgovdrupal/localgov_microsites_group/issues/74#issuecomment-1134792859


finnlewis (approved):
Neat!

PR #85, opened 2022-05-24
Add and update help text

msayoung:


config/optional/core.entity_form_display.group.microsite.default.yml
```diff
@@ -372,7 +369,7 @@ third_party_settings:
         description: ""
         required_fields: true
 _core:
-  default_config_hash: c43_05rkGWfTKnr8L-DPfY4zz8CY8Fimrs-c9df8ejM
+  default_config_hash: Zb2vo5iedq8-q2_y5V9hzeckbrXhbvMFdw5mB7T-oz4
```
ekes: This probably doesn't want to be here?

ekes (changes requested):
Other than the one hash, text changes make sense; and the one view empty text.
Still don't know why we're switching between ' and " it would be good to know as it makes reading the changes more difficult.

config/optional/core.entity_form_display.group.microsite.default.yml
msayoung: It's gone!

ekes (approved):

Issue #82, opened 2022-05-24
show the Tours automatically to people on their first visit


finnlewis:

stephen-cox:
@msayoung @finnlewis It seems that there's not a keen interest in having Tour autoplay to people when they first visit a page: https://www.drupal.org/project/drupal/issues/2934417

The agreed autoplay method is to add `?tour=1` to the end of the URL. For example: https://microsites.localgovdrupal.org/admin/microsites?tour=1

I'm sure we could get the tours to autoplay when first visiting the page, but the code won't be trivial as we'll need to set a cookie to ensure it only plays once.

Can we use the `?tour=1` somehow when getting people to visit pages for the first time?

As it stands the controller user doesn't have permissions to use Tour. This PR #86 fixes this and some other issues I found while looking into this.

finnlewis:
Still a live one?

PR #86, opened 2022-05-24
Update Tour config

stephen-cox:
From #82 

- Moves tour config from install to optional.
- Corrects the dependency declarations in the info.yml for for core modules.
- Adds permission to use Tour to the LocalGov Controller user

### DIFF
--- a/config/install/user.role.microsites_controller.yml
+++ b/config/install/user.role.microsites_controller.yml
@@ -6,6 +6,7 @@ dependencies:
     - group
     - system
     - toolbar
+    - tour
 id: microsites_controller
 label: 'Microsites controller'
 weight: 2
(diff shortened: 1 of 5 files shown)


ekes:
Looks fine. Why the move to optional though. Seeing as user (install) is still dependent on tour?

stephen-cox:
> Why the move to optional though. Seeing as user (install) is still dependent on tour?

Having config in optional is recommended in the docs (https://www.drupal.org/node/1934442), but giving we have an explicit dependency it can be in install. Happy either way

ekes:
Guess that makes sense for tour, and might make it easier to unpick later, if we want to move stuff into submodules. 
It's not like our wonderful media and group config in optional :)

ekes (approved):

PR #95, opened 2022-05-25
Invalidate node cache when group is updated

### LINKED ISSUE #94: Clear node group content cache when saving a group
When adding footer blocks (on council test sites) they can't be seen immediately. They turn up after a while, or after a cache clear

stephen-cox:


ekes:
Isn't there a domin:$id cache tag? Might be better?

localgov_microsites_group.module
```diff
@@ -71,6 +72,15 @@ function localgov_microsites_group_entity_extra_field_info() {
     ->entityExtraFieldInfo();
 }
 
+/**
+ * Implements hook_ENTITY_TYPE_presave().
+ */
+function localgov_microsites_group_presave(GroupInterface $entity) {
```
finnlewis: Is this right? 

If the name of the module is "localgov_microsites_group"

Should the hook not be: 

localgov_microsites_group_**group**_presave() 

I was also trying to create a similar hook to hook into the save and set an additional message, but could not get it to fire.

I also see above we have: 

```
/**
 * Implements hook_ENTITY_TYPE_view().
 */
function localgov_microsites_group_view(
```

I wonder if that one is firing.

stephen-cox:
> Isn't there a domin:$id cache tag? Might be better?

When looking at the cache tags for a group node they had `node:2` and `node_view`. `node:$id` is no good so I went with `node_view`. There aren't any domain specific ones and group ones didn't work.

This is needed for testing tomorrow, so as long as it works I suggest it's merged. Maybe we should implement some domain/group cache tags for this sort of thing?

finnlewis:
Either way, this does appear to fix the caching issue. 

Had to reconfigure my local to actually get it to cache (settings.lando.php) and then replicated the odd behaviour.

Checking out this branch made it play nice again!

Happy with this for now for testing and we can make it slicker soon.

finnlewis (approved):
Works for now!

localgov_microsites_group.module
stephen-cox: Nice catch. I was using hook_presave() not hook_ENTITY_TYPE_presave(). I have updated the function name and it still works as expected.

Looks like the localgov_microsites_group_view function name should be update too, but I haven't done that here.

stephen-cox:
I have merged this, but will leave #94 open as we need a better solution for this in the long run.

Issue #96, opened 2022-05-25
Menus - WSOD on node/add


finnlewis:
Creating a new node, adding a menu item creates a WSOD

Further to this, editing node, no menu widget.

Options: 
 - fix
 - remove menu widget from form

finnlewis:
I think this was done @stephen-cox - closing - feel free to re-open if not!

PR #97, opened 2022-05-25
Temporarily hide menu widget on node creation pages 

stephen-cox:
Problem on #96 still needs fixing, but this hides the issue while user testing is going on


msayoung (approved):
gorgeous

PR #100, opened 2022-06-13
Feature/1.x/improve frontend options

msayoung:


finnlewis (approved):
Installs fine, and config style is better again, so let's merge it for further review on the test site

PR #104, opened 2022-06-17
Add, remove and reorder font options

msayoung:


finnlewis (approved):
Looks safe enough to me!

Issue #105, opened 2022-06-20
Move all group permissions to localgov_microsites_permissions module


stephen-cox:
Group permissions need some unpicking because they are creating dependencies across everything.

At a minimum we should move the permissions config files to the localgov_microsites_permissions module, but maybe we want to implement something like what we're doing with `hook_localgov_roles_default`

Thoughts @ekes @finnlewis ?

finnlewis:
I'm not clear how putting all permissions in a separate module helps. Could we discuss this at some point or clarify @stephen-cox @ekes ?

stephen-cox:
@finnlewis @ekes Rather than moving permissions to localgov_microsites_permissions (which I think can be removed), I propose adding a RolesHelper class like (https://github.com/localgovdrupal/localgov_core/blob/2.x/modules/localgov_roles/src/RolesHelper.php) and using this to decouple the dependency between localgov_microsites_group and any modules that need to define permissions.

PR with the proposal following shortly.

PR #106, opened 2022-06-20
Add Directories to Microsites

### LINKED ISSUE #57: Site 1 admin - create directories in microsite 1
## User story

As a 'site admin' on microsite
I want to create directory channels and directory entries 
So that we can deploy a searchable directory of locations on a microsite.

## Notes: 

We think this might actually be easier than adding localgov_events into the microsites install.

## Possible complications: 
 - seach_api with group content -> which are we showing, the group content relationship or the actual node with the content
 - access control on the node might help with the above

## Assumptions: 

We are adding localgov_directories as is to microsites, no change in functionality.

We will separate off the issues of: 
 - restricting which content types are available to each microsites
 - we allow all facet types to be accessible across all microsites: later we might want to restrict these per group

## Acceptance tests: 

Setup: controller: 
 - creates a new microsite 'microsite 1'
 - creates and assigns a new user to the site as a microsite admin

As 'microsite 1 admin'

1. Log into site
2. Create directory channel
3. Create several directory entries
4. Log out
5. Search
6. Filter
7. Smile

stephen-cox:
This PR creates a new localgov_microsites_group_content module for group content, moves Page configuration into it and adds Directories support.

There are some issues still to resolve:

- [ ] Directory facets are configured as group content, but directory facet types aren't. It shouldn't be too hard to make directory facet types group content, but wanted to get something up for testing for sprint review. (https://github.com/localgovdrupal/localgov_microsites_group/issues/110)
- [ ] Directory venues aren't enabled by default because the Drupal site install hangs indefinitely when trying to enable something to do with this. It is possible to enable venues after a site install, so I suspect some sort of run condition, but haven't spent much time investigating this yet. (https://github.com/localgovdrupal/localgov_microsites_group/issues/108) 
- [ ] When venues are enabled it's not possible to add addresses as anyone other than User 1. I expect something like https://github.com/localgovdrupal/localgov_microsites/issues/111 is going on. (https://github.com/localgovdrupal/localgov_microsites_group/issues/109)
- [ ] There are errors when enabling directories that don't effect functionality, but would be good to fix: https://github.com/localgovdrupal/localgov_directories/issues/194
- [ ] Permissions are a bit of a mess as they are defined in localgov_microsites_group. I suggest moving them to a new module #105 
- [x] **Needs tests**

Fixes localgovdrupal/localgov_microsites#57


ekes (commented):
Manually tested; created facet types; directories; added pages.
Adding the facet types as GroupContent makes sense; and the code for controller looks good.
Tests would maybe throw up some points.

Thought: Would it be better to split these out into more modules with hard dependencies on the appropriate content types? Then the config can be install rather than optional. There's also less unnecessary code enabled. And possibly down the line it'll be clearer what each bit is for. Could be particularly useful with all the extra controllers and group related code per entity type.
[this is more a merge, refactor, and test, thought; not a do this first thought]

stephen-cox:
@ekes Added tests which are all passing. Good for you to review now.

finnlewis:
I'm just taking this for a spin locally too.

finnlewis:
Installs OK for me, I can run the tests and they pass. Nice to see how it's done on different domains.

Setting up a directory channel and facets works as a microsite admin on site 1, and I can create content and see it listed. 

For some reason I don't see the facets on the directory channel: 

image

Is that just block placement config that has not installed for some reason? 

Generally good though, would be good to get this merged in so @markconroy and @msayoung can take a look at the styling.

stephen-cox:
> Is that just block placement config that has not installed for some reason?

Does look like the blocks haven't been placed for you. Not sure why though as there's config in the PR to place them as long as the Microsites Base theme is enabled. (We may want to move the block config to the theme as we've done elsewhere).

stephen-cox:
Once (if) PR #125 gets merged the permissions defined in the `localgov_microsites_group` module config can be moved to the new `localgov_microsites_directories` module, breaking the circular dependency between the two modules.

ekes (commented):
We probably need to add to the composer.json for this module:
```
require-dev {
    "localgovdrupal/localgov_directories": "2.x-dev"
},
suggests {
    "localgovdrupal/localgov_directories": "For directories content type in microsites"
}
With the profile separately requiring it.

File: modules/localgov_microsites_directories/localgov_microsites_directories.module [hunk 1]
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * Implements hook_entity_insert().
+ */
+function localgov_microsites_directories_entity_insert(EntityInterface $entity) {
+
+  if ($entity->getEntityTypeId() != 'localgov_directories_facets') {
+    return;
+  }
+
+  $group = \Drupal::service('group.group_route_context')->getGroupFromRoute();
```
ekes, line 21: Just confirming we are _always_ in the route group context when directory facets get inserted? (Here and below a couple of times)
Otherwise could checking the active domain group if there isn't one returned be worth doing?

ekes (approved):
Just the idea about the composer.json; and the early blocks noted by Finn.
And a question over if the context can always be relied on existing -- I think it can actually.

With the blocks and permissions being sorted. I'd say it's just the composer.json
For the rest looks good.

modules/localgov_microsites_directories/localgov_microsites_directories.module
stephen-cox: As things stand the group ID is always in the route when editing a group. This applies to both the controller site and when on the site itself. E.g. https://microsites.localgovdrupal.org/group/1/edit and https://site1.microsites.localgovdrupal.org/group/1/edit. So this is fine for the moment.

Is there chance this behavior will change?

stephen-cox:
> We probably need to add to the composer.json for this module:
> 
> ```
> require-dev {
>     "localgovdrupal/localgov_directories": "2.x-dev"
> },
> suggests {
>     "localgovdrupal/localgov_directories": "For directories content type in microsites"
> }
> ```

Yep - good idea - done

stephen-cox:
Just trying to reproduce the missing blocks.

It seems to work if I do the following:

```
composer require localgovdrupal/localgov_microsites:"dev-feature/57-directores as 1.x-dev" -W
composer require localgovdrupal/localgov_microsites_group:"dev-feature/57-directories as 1.x-dev" -W
drush si localgov_microsites
```

@finnlewis Can you provide steps on how you installed stuff without the blocks?

finnlewis:
@stephen-cox following your steps the blocks are just fine and dandy, thanks dude!

finnlewis (approved):
Let's get this in so we can review styling and make it beautiful!

PR #117, opened 2022-06-24
Use group menu block from original module with new context.

ekes:
localgovdrupal/localgov_microsites#156

Revert "Add block config for main and utility menus (#41)"
This reverts commit 392940fecf61c8d685564ca59cd0b0c6f95b278c.


ekes:
Requires https://github.com/localgovdrupal/localgov_microsites_base/pull/55 before merging.

stephen-cox (approved):
Yep - no need for this once the menu blocks are using the new context

PR #119, opened 2022-06-27
Don't use two step for adding Node Page GroupContent.

ekes:
No additional fields. Alias will be set with domain_path see #146


ekes:
[Got confused by the fact that the page has a required field - the paragraph type - but it doesn't display a very clear error if there isn't a paragraph added. It looks like the only required fields are title and summary.]

stephen-cox (approved):
This definitely removes the 2nd step.

The 2nd step was the only way to add a path alias, is this moving to the node form now?

ekes:
Moves to node (or other content entity) form with domain_path module to allow a path per domain localgovdrupal/localgov_microsites#146

PR #124, opened 2022-06-29
Add an image style for the footer logos

msayoung:


config/optional/core.entity_view_display.group.microsite.microsite.yml
```diff
@@ -111,8 +112,12 @@ content:
     label: hidden
     settings:
       image_link: ''
-      image_style: ''
-    third_party_settings: {  }
+      image_style: localgov_microsite_200_85
```
markconroy, line 115: I _think_ it might be better to call this image style something generic, like 'Microsite footer logo' in case we want to change the dimensions of it later.

msayoung: Most of the others in LGD are named after their size or aspect ratio so I think it makes sense to keep it like this.

markconroy: Let's merge this mother so!

PR #125, opened 2022-06-29
Add microsite_admin role and HOOK_localgov_microsites_roles_default to allow modules to define permissions without creating a dependency on localgov_microsites_group

### LINKED ISSUE #105: Move all group permissions to localgov_microsites_permissions module

### LINKED ISSUE #144: Create `microsite admin` role
We currently need to give authenticated users quite a few permissions to do stuff (create media, manage path aliases, manage directory facets). It's not ideal adding these permissions to the authenticated user role as it prevents the creation of other user types that shouldn't have these permissions.

We should create a new user role (microsite admin?) and assign these sorts of global permissions to them.

@ekes @finnlewis Thoughts on this? Can it be done with group permissions or will there always be a need to assign some global permissions to site admins?

stephen-cox:
This PR creates a `microsite_admin` role and moves permissions from the authenticates user to this and the `microsite_controller`.

This is also creates a hook to allow modules to define their own permissions. meaning we can remove any dependencies between this module and the other module (while keeping a dependency between the other module and this one).

Fixes #105 
Fixes localgovdrupal/localgov_microsites#144


finnlewis (approved):
Works for me @stephen-cox, thanks!

Only one niggling question about naming: we define the 'microsite admin' user as distinct from the 'microsite editor' user within a microsite. Unless I'm mistaken, this microsites_admin role will be given to all users of microsites, so might this cause confusion down the line? If so, we could call it microsites_user or something that does not clash. 

Am I right in thinking that the permissions are then restricted on the group by group role assigned (admin / editor etc)

stephen-cox:
> Only one niggling question about naming: we define the 'microsite admin' user as distinct from the 'microsite editor' user within a microsite. Unless I'm mistaken, this microsites_admin role will be given to all users of microsites, so might this cause confusion down the line? If so, we could call it microsites_user or something that does not clash.

You're right in there is there is a potential for confusion here. The `microsites_admin` role is a global role (the `microsite_admin` role - no 's' - is a group role). I did wonder what to call it.

The `microsites_admin` role is given to anyone who needs to edit content on a microsite - currently required for media, directory facets and aliases. Perhaps `microsites_editor` is better? We don't have an editor role of any sort yet.

ekes:
Which users of the site who can log in _don't_ need to have these permissions? Maybe that isolates what type, name, of user this is?

stephen-cox:
> Which users of the site who can log in _don't_ need to have these permissions?

At the moment no one; which is why the permissions have been added to the authenticated user role, but this means we can't restrict who can do what later on.

I suggested an admin role, because this gives the flexibility to create roles like directory editor later on.

stephen-cox:
I have rename the microsites_admin role to microsites_trusted_editor. Is this okay for now?

finnlewis:
Aye - let's go with that for now! Thanks @stephen-cox

finnlewis (approved):

PR #127, opened 2022-06-29
Adds fields to override header and footer styles

msayoung:


finnlewis (approved):
Installs fine and I see the fields, thanks @msayoung

Issue #107, opened 2022-06-21
Redirect users from group content entities to the content


finnlewis:
Once we've sorted the GroupContent question (https://github.com/localgovdrupal/localgov_microsites_group/issues/23)

We'll probably need to hide access to the group content entity and redirect users from said entity to the actual content, just in case!

ekes:
PR https://github.com/localgovdrupal/localgov_microsites_group/pull/129

ekes:
Moved to done. Closing.

PR #129, opened 2022-07-04
Redirect canonical (for view) to the content entity #107

ekes:
Feel like making 'edit' and 'canonical' would be nicer to remove 'view'
completely, but there seem to be plenty of assumptions it's there, so
this is quick.

### DIFF
--- a/config/optional/views.view.group_nodes.yml
+++ b/config/optional/views.view.group_nodes.yml
@@ -1,8 +1,6 @@
 langcode: en
 status: true
 dependencies:
-  config:
-    - group.type.microsite
   module:
     - gnode
     - group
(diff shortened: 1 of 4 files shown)


ekes:
Question.

This works for all content types. Maybe it should only be Node?

Some relations don't have view already. This is better.

The existing one that does have view is 'User', maybe being able to view this rather than being redirected to the user is desirable?

ekes:
> This works for all content types. Maybe it should only be Node?
> Some relations don't have view already. This is better.

What is it that stops 'Content Menu' and 'Directories Facet' GroupContent from having a visible canonical route, but still have the permission to be viewed by users? Would this be better for Node GroupContent - a 403 and never a link, rather than a redirect?

msayoung:
@stephen-cox @finnlewis @msayoung @markconroy thoughts?

stephen-cox:
@ekes The group node view works and node CRUD operations also work fine, but I have to admit that I'm not entirely sure what I should be testing here

ekes:
> @ekes The group node view works and node CRUD operations also work fine, but I have to admit that I'm not entirely sure what I should be testing here

The PR redirect the canonical view of any GroupContent to the content itself. So /group/1/content/1 just redirects to /node/1 rather than showing a blank page.

My question was if that's actually desirable for Users; and if we couldn't rather do what Group Content Menu and others do to make there be no GroupContent 'view' path at all, so you never seem to get 'view' as an option for their GroupContent entities, just edit (I'm not sure how they're doing that).

ekes:
Proposal https://github.com/localgovdrupal/localgov_microsites_group/issues/23#issuecomment-1187019297 merge this; then make issues for better / content type specific changes?

finnlewis (approved):
Looks good to me @ekes

ekes:
OK lets go with this and refine if need be.

PR #130, opened 2022-07-07
Adds config for ginvite : group invitations

msayoung:


msayoung:
Questions @finnlewis 

Do we want to change the word _group_ to the words _microsite_ in the invitations.? (and probably generally).
Do we want the links in the emails to point to the microsite rather than the control site? (currently it defaults to whichever site the invitation was generated from)

finnlewis:
> Do we want to change the word group to the words microsite in the invitations.? (and probably generally).

Yes, I think that sounds like a great idea!

> Do we want the links in the emails to point to the microsite rather than the control site? (currently it defaults to whichever site the invitation was generated from)

I think defaulting to whichever site the invitation is generated makes sense. I imagine that usually this will be from the microsite.

finnlewis:
@msayoung still some merge conflicts on this one. Could you take a look as I'm not sure which way it's meant to be.

finnlewis (approved):
Tested the functionality locally and it works. Invite sent, register for an account, confirm email address, able to log into the site. 

As you mention @msayoung we will need to look at the wording of the emails to ensure the next steps are clear, but we can do that on another PR I reckon.

If we can resolve the merge conflicts with the permissions I think this is good to go.

msayoung:
@finnlewis merge conflict resolved, but I dont know whether the directory content type permissions are correct. Not sure how they got in.

finnlewis:
Some coding standards to fix @msayoung https://github.com/localgovdrupal/localgov_microsites_group/runs/7405601612?check_suite_focus=true#step:4:19

msayoung:
The tests aren't working yet @finnlewis  so we need to either take them out or finish them

finnlewis:
Testing locally I get some odd behavior

Microsite 1 has admin1 as a group admin
Microsite 2 has admin2 as a group admin

admin1 logs into microsite 1 and can see group invitations at 
https://localgov-micro-1.lndo.site/group/1/invitations

Now expecting that admin1 cannot do the same on microsite2

BUT
admin1 logs into microsite 2 and can see group invitations at 

https://localgov-micro-2.lndo.site/group/1/invitations

and can also see: 
https://localgov-micro-2.lndo.site/group/1/domain-settings
https://localgov-micro-2.lndo.site/group/1/members
https://localgov-micro-2.lndo.site/group/1/nodes
https://localgov-micro-2.lndo.site/group/1/menus
https://localgov-micro-2.lndo.site/group/1/directory-facets

BUT when we go to 
https://localgov-micro-2.lndo.site/group/1 

we get 403 ACCESS DENIED

ekes:
So we discussed this one to get in and add the tests in a separate issue/PR. Checking with @stephen-cox but I'd like to merge this and then merge it into #160 which has the trait for helping create microsites groups to extend for this. For logging-in testing I think that should be a separate issue (probably on domain_group).

stephen-cox (approved):
Let's merge it

Issue #136, opened 2022-07-11
Managing vocabularies and taxonomy terms in a group


stephen-cox:
Current plan is for vocabularies to be shared across groups and taxonomy terms to be group content.

We will need to implement a UI for this. One's been built for managing directory facets, so something similar can be done for this.

stephen-cox:
@ekes This will be required for both news and events content types, so have created a separate issue for it.

Is there anything else we need to consider here?

ekes:
Don't know if we want to look at https://www.drupal.org/project/group_taxonomy

I'm pretty sure we don't want to use the patch that would add vocabulary (see issue about config entities elsewhere); but maybe it's better to use it for the terms? Or maybe it's better doing it ourselves the same?

ekes:
My other probable requirement is we will want to differentiate between vocabularies:- I can see the use case where Open Active vocabulary or such has been imported and is fixed, so that vocabulary has terms shared between all sites that can't be edited; where as another can.

ekes:
Oh.

https://www.drupal.org/project/group_term

stephen-cox:
I had a look at both those.

I don't think we want [Group Taxonomy](https://www.drupal.org/project/group_taxonomy) as we need that patch. At the moment people seem okay with shared vocabs and group terms.

[Group Term](https://www.drupal.org/project/group_term) will make it easier to create group content plugins as we won't need to write our our plugin classes for each vocab, but that it pretty much all it provides. We would still need to create a UI for managing group terms. One option would be to pacth/fork the module to add a UI. Only concerns with Group Term is it alpha without much dev and only used on 55 sites.

ekes:
Had a look now. Three commits. I'm guessing because it is really just the enabler, view, tokens and permissions. Most of which seems c&p code.

But if I'm right it is breaking up the permissions per bundle and the ContentEntabler, which is good.

So if I read it correctly we could use that, save us doing the same c&p. If there are issues down the line that are a problem we can fork - or ask for maintainership?

Then make a localgov_microsites_group_term_ui that adds the UI for managing the terms? Rather than try and get it into group_term?

stephen-cox:
Sounds like a plan. Thoughts on whether localgov_microsites_group_term_ui module should be standalone rather than a submodule of localgov_microsites_group? I lean towards a standalone module, but it's going to be quite simple so could just go into localgov_microsites_group.

ekes:
[...]

PR #137, opened 2022-07-12
Issue localgovdrupal/localgov_microsites#118 Initial import of localgov_news into groups.

ekes:
Tests pass :crossed_fingers: 
Terms not yet in the groups, separate issue #136

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,10 +13,12 @@
     "localgovdrupal/localgov_core": "^2.1.0"
   },
   "require-dev": {
-    "localgovdrupal/localgov_directories": "^2.0"
+    "localgovdrupal/localgov_directories": "^2.0",
+    "localgovdrupal/localgov_news": "^2.1"
   },
   "suggests": {
-    "localgovdrupal/localgov_directories": "For directories content type in microsites"
+    "localgovdrupal/localgov_directories": "For directories content type in microsites",
+    "localgovdrupal/localgov_news": "For news content type in microsites"
   },
   "extra": {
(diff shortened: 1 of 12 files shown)


stephen-cox:
With the inclusion of news in the profile (https://github.com/localgovdrupal/localgov_microsites/pull/188) tests should be passing, but there's an issue with a missing dependency on the link_attributes module.

ekes:
Hm... https://github.com/search?q=org%3Alocalgovdrupal+link_attributes&type=code and
https://github.com/localgovdrupal/localgov_news/blob/2.x/localgov_news.info.yml#L21=
but not https://github.com/localgovdrupal/localgov_news/blob/2.x/composer.json

ekes:
We should get localgovdrupal/localgov_news#63 merged Monday.

I've also now noticed that:

Newsroom creates a /news node by default on installation.
It gets in there before our default content.
/node/1 becomes a newsroom node.
When you create a microsite it's coded to clone /node/1 by default.
BOOM!

Suggested solutions?
 - When setting the default content configuration lookup the content by title?
 - In any case add in some checks for content type before creating the GroupContent.
 - Something else?
 
@stephen-cox any ideas?

ekes:
Dependencies in LocalGov News https://github.com/localgovdrupal/localgov_news/pull/63 will need a release.

finnlewis (approved):
Tested locally, installs fine and I can create news rooms and news on each microsite. Good to get this in and tidy up the styling next sprint I reckon.

PR #138, opened 2022-07-12
Add Events to microsites

stephen-cox:
Part of https://github.com/localgovdrupal/localgov_microsites/issues/119

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,10 +13,12 @@
     "localgovdrupal/localgov_core": "^2.1.0"
   },
   "require-dev": {
-    "localgovdrupal/localgov_directories": "^2.0"
+    "localgovdrupal/localgov_directories": "^2.0",
+    "localgovdrupal/localgov_events": "^2.0"
   },
   "suggests": {
-    "localgovdrupal/localgov_directories": "For directories content type in microsites"
+    "localgovdrupal/localgov_directories": "For directories content type in microsites",
+    "localgovdrupal/localgov_events": "For events content type in microsites"
   },
   "extra": {
     "enable-patching": true,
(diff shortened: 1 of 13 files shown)


stephen-cox:
Tests are failing due to not pulling in the LocalGov Events module; they pass on my local setup.

Doesn't do anything with facets as this needs sorting first #136 

There's an issue with blocks being placed on the /events listing page when they shouldn't be there. See #139

ekes:
I was wondering if we shouldn't set access to the /events page based on the events content type being available to the group? We've not made the page for the microsite admin to configure this yet, but it would then be removing the access to view (and create) the event content type in the group, so the check could be on that group permission?

stephen-cox:
> I was wondering if we shouldn't set access to the /events page based on the events content type being available to the group? We've not made the page for the microsite admin to configure this yet, but it would then be removing the access to view (and create) the event content type in the group, so the check could be on that group permission?

I did wonder the same and probably makes more sense; there are times when you might want to show an empty events view.

The ability to enable content types on a site by site basis doesn't exist yet. Do you think the current solution of restricting access to /events is acceptable until then?

ekes:
> > I was wondering if we shouldn't set access to the /events page based on the events content type being available to the group? We've not made the page for the microsite admin to configure this yet, but it would then be removing the access to view (and create) the event content type in the group, so the check could be on that group permission?
> 
> I did wonder the same and probably makes more sense; there are times when you might want to show an empty events view.
> 
> The ability to enable content types on a site by site basis doesn't exist yet. Do you think the current solution of restricting access to /events is acceptable until then?

Yep added to #14

finnlewis (approved):
Yeah baby - seems to work well for me locally. Events are created on different sites and are only visible on the site in question. Tests pass.. let's get this in and make it beautiful!

PR #140, opened 2022-07-14
Add fields to customise the appearance of the menu

msayoung:


ekes (approved):
More strings and options fields, moving two existing, into a fieldset for configuring menus. If I understand? Then that looks fine.

PR #141, opened 2022-07-15
Domain Group settings form

ekes:
Working with https://git.drupalcode.org/issue/3278349-3293374/-/merge_requests/1
Makes changes to work with the now two form plugins.
Adds permissions so that controller can access all of the form on all groups, and site admin can access the site settings part of the form localgovdrupal/localgov_microsites#157

### DIFF
--- a/config/install/group.role.microsite-admin.yml
+++ b/config/install/group.role.microsite-admin.yml
@@ -13,6 +13,7 @@ permissions_ui: true
 permissions:
   - 'access group content menu overview'
   - 'access group_node overview'
+  - 'administer group domain site settings'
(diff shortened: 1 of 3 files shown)


ekes:
I'll merge https://git.drupalcode.org/issue/3278349-3293374/-/merge_requests/1 at the same time as this. Otherwise the form is broken when only one of them is done.

finnlewis (approved):
Tested locally with the branch from https://git.drupalcode.org/issue/3278349-3293374/-/tree/feature/3.x/3293374-domain-config-plus-permissions and it works as expected.

PR #142, opened 2022-07-15
Created localgov_microsites_group_term_ui module

### LINKED ISSUE #136: Managing vocabularies and taxonomy terms in a group

stephen-cox:
Fixes #136

~~Needs tests~~


stephen-cox:
@ekes @finnlewis This PR adds a UI for managing taxonomy terms as group content and configures news and events for the vocabs used by them. It's worth a check now to ensure it all makes senses.

There is a major issue that prevents this from being merged. Entity references fields on nodes list all the terms, not just the group ones. Looks like we'll need to ensure these entity reference fields are filtered for group content. There are a number of ways I can think of to solve this, but they all involve making config changes or writing code specific for each field.

@ekes Are you aware of a way that we can check access to all terms when listed in an entity reference field so they can be filtered accordingly?

ekes:
> @ekes Are you aware of a way that we can check access to all terms when listed in an entity reference field so they can be filtered accordingly?

I was assuming it would need a custom widget; but then I was thinking about filtering the vocabularies that weren't in use.
For access to _*view*_ the title of a term, if the user doesn't have view permissions they shouldn't see it I think. Maybe that's an issue with the way the query is altered for entity access.

stephen-cox:
Further investigation into why terms are not being filtered by group content shows that term selection widget directly loads the Terms for a Vocabulary, rather than using an entity query, which is the default.

See: https://git.drupalcode.org/project/drupal/-/blob/9.5.x/core/modules/taxonomy/src/Plugin/EntityReferenceSelection/TermSelection.php#L52

Compared to: https://git.drupalcode.org/project/drupal/-/blob/9.5.x/core/lib/Drupal/Core/Entity/Plugin/EntityReferenceSelection/DefaultSelection.php#L337

Maybe the cleanest solution is write our own EntityReferenceSelection plugin (possibly extending TermSelection) and then update our field config to use this.

finnlewis:
Interesting @stephen-cox, good detective work Inspector Cox! 

Should we mark this PR as draft and write the plugin?

ekes:
Yer an entityquery gets altered, ::loadTree uses a direct select query on the table. It does then iterate over the loaded terms and remove any that aren't published if the user doesn't have access (it also then importantly removes any children); the best future-proof option might be to inherit this and do the filter again in the same way, but checking access.

stephen-cox:
@ekes This has turned into a bit of a monster, but is ready for review now.

modules/localgov_microsites_group_term_ui/src/Plugin/EntityReferenceSelection/GroupTermSelection.php
```diff
@@ -0,0 +1,114 @@
...
+    $group_id = \Drupal::service('domain_group_resolver')->getActiveDomainGroupId();
+    if ($group_id) {
+      $group = \Drupal::entityTypeManager()->getStorage('group')->load($group_id);
+    }
+    if (empty($group)) {
+      $group = \Drupal::request()->attributes->get('group');
+    }
+
+    // Only show group terms if there's a group.
+    if ($group instanceof GroupInterface) {
```
ekes, line 100: Guess rewriting and changing the access check of the parent wasn't an/the best option?
Bit of an edge issue but I think the parent deals with removing the hierarchy if the term above isn't visible. Not sure what's going to happen here. Maybe we just make a follow-up.

stephen-cox: Just adding an access check didn't work. Not sure why, but the check always came back TRUE. Once we started loading the group it seemed easier to call the parent and then filter that.

Good call about the hierarchy. The UI for managing group terms doesn't actually deal with the hierarchy as it's a flat view. although it is possible to add a parent term on the term edit form.

ekes: Follow-up https://github.com/localgovdrupal/localgov_microsites_group/issues/176

ekes (approved):
Follow-up issue opened.

Rest works with manual testing, and tests pass. Let's get it in for User Testing.

PR #135, opened 2022-07-11
Enable directory venues as part of a site install 

stephen-cox:
~~Blocked by https://github.com/localgovdrupal/localgov_directories/pull/202~~

Fixes #108

PR #148, opened 2022-07-21
Add promo page to the localgov_microsites_group

msayoung:


finnlewis:
Maria would like some other eyes on this, it's not working for her locally.

msayoung:
When I run a site install this just hangs forever, I'm not sure what is causing this. Oddly the tests are passing here.  @finnlewis can you replicate?

markconroy:
Hanging for me too.

But ... I think I am noticing something, or maybe not.

The branch for this seems to be cut from the `localgov_microsites_group` repo, but the code it targets is in the `localgov_microsites_directories` module. Am I right in that thinking? If so, we probably need to more this PR to the directories repo.

msayoung:
I was following the pattern we have for bringing news an events in to microsites. A submodule within localgov_microsites_group https://github.com/localgovdrupal/localgov_microsites_group/tree/1.x/modules

markconroy:
Sorry, that's my mistake. I was reading the folder names incorrectly. 

So, it looks like something in this is causing the install to hang indefinitely, but no error reported. I've no idea what that might be.

stephen-cox:
This is likely to be hanging for the same reason venues fails; https://github.com/localgovdrupal/localgov_directories/issues/201

stephen-cox:
This should work once #135 is merged.

msayoung:
This works for me locally now! Whoop! Thanks @stephen-cox 
@finnlewis wanna give it a whirl and then we can merge?

stephen-cox (changes requested):
Worked fine apart from the permissions weren't being installed. 

Looks like you should be using `group_node:localgov_directory_promo_page` in the permission name instead of `group_node:localgov_directories_promo_page`

markconroy:
Looks good to me (once the permission issue @stephen-cox mentioned is sorted). Ready to merge then.

stephen-cox (approved):
Looks good now

PR #149, opened 2022-07-21
Adds the layout paragraph to news content type

msayoung:
@stephen-cox - here's the config for adding the layout paragraph field to news, but as predicted we cant install the core.entity_*_display files as they already exist. 

Can you take it from here?

PR #156, opened 2022-07-25
adds config for events nodes in full view mode

### LINKED ISSUE #155: Manage display settings for event nodes

markconroy:
Closes #155


msayoung:
I think it's missing the  `config/optional/field.storage.node.lgms_banner.yml ` which we may find useful across our content types.

With that file then the field is added to the content type, but the display modes etc aren't updated as the active configuration is prioritised. @stephen-cox was going to write an update hook on this PR https://github.com/localgovdrupal/localgov_microsites_group/pull/149 to do something simialr, so hopefully we can reuse it here.

msayoung:
Sorry, just seen this pr with the lgms_banner config https://github.com/localgovdrupal/localgov_microsites_group/pull/154

stephen-cox:
As @msayoung points out, this isn't going to work as expected. The config is in optional and as config settings with the same name already exists the config will never be imported. Moving it to install (where it probably should live) is going to throw an error.

Any changes to existing config need to be done in and install or update hook. This is how I've implemented it for the paragraphs field on events: https://github.com/localgovdrupal/localgov_microsites_group/blob/eb1ec0ade5294e018f6e4c28840e5e821394a5c5/modules/localgov_microsites_news/localgov_microsites_news.install

markconroy:
@stephen-cox I've removed the config for the form/view displays now, so this PR _only_ creates the banner field for the events content type.

Let's get this merged and we can follow-up with programmatically adding the displays in our `hook_install()`

msayoung (approved):

stephen-cox (approved):
Looks good :+1:

PR #157, opened 2022-07-25
Allow trusted editors to use responsive preview and admin toolbar

msayoung:


msayoung:
@stephen-cox I've added responsive preview permissions, and the use toolbar permission to the trusted editor. 

With this they can preview (responsively or normally) content which has already been saved.

finnlewis:
Trying to work out why the MicrositeDirectoryContentTest test is failing.

https://github.com/localgovdrupal/localgov_microsites_group/runs/7519565353?check_suite_focus=true#step:5:22

finnlewis:
aha! 

``` Cannot search on a disabled index. ```

**Comment by stephen-cox** [comment]
Tests should be fixed with #135

**Comment by stephen-cox** [comment]
From the looks of things the responsive_preview module is required by composer in the localgov_page module.

https://github.com/localgovdrupal/localgov_page/blob/7527d2e47aa88d3e2db2d0a7e7f92b71448e4b74/composer.json#L10

It should be removed from there and added here.

**Review by stephen-cox — CHANGES_REQUESTED** [review]
Need to add a dependency to composer.json for 

```
"drupal/responsive_preview": "^1.1",
```

msayoung:
Done here and https://github.com/localgovdrupal/localgov_page/pull/22

stephen-cox (approved):
Works fine, but I wonder if we should be giving site admins access to the toolbar, given there's nothing else in it and all site management links are at the top of the page. Happy to include it for now if this is something we want to test.

PR #160, opened 2022-07-26
Add service to enable / disable permissions per module.

ekes:
To be used by domain settings form to "enable/disable" modules per domain.


ekes:
@todo The form

ekes:
@stephen-cox Shall we just remove the kernel test?

stephen-cox:
> Shall we just remove the kernel test?

Hopefully fixed :crossed_fingers:

ekes:
It's green!

stephen-cox:
Looks to work fine apart from some weirdness on the content type setting form. Things can be enabled, but the form shows is greyed out without a check so saving the form disables it.

Example
image

ekes:
Tests require https://github.com/localgovdrupal/localgov_paragraphs/pull/72 (plus probably more) to fix the schema errors. Rest of tests should pass.

finnlewis:
@stephen-cox is happy to bring this in now,

PR #163, opened 2022-07-26
Update .install file to add the new banner field to the view and form…

msayoung:
… display


msayoung:
@markconroy this one also needs the field groups added to the view display and then the placement of the fields.

markconroy:
I think this is ready now for review. 

1. Creates view display field group for "Banner"
2. Creates view display field group for "Page Top"
3. Sets paragraphs field into "Story" tab on node edit form

stephen-cox (approved):
Looks good

PR #166, opened 2022-07-28
Move field config from optional to install

### LINKED ISSUE #164: Move field config from `optional` to `install`
There's a bunch of field configuration in the optional config directory and other config in install config directory that relies on this config. This is causing unexpected dependency issues during a site install.

I'm currently getting 

```
In UnmetDependenciesException.php line 100:
                                                                                                                                                                                                                                
  Configuration objects provided by <em class="placeholder">localgov_microsites_events</em> have unmet dependencies: <em class="placeholder">field.field.node.localgov_event.lgms_banner (field.storage.node.lgms_banner)</em>                                                                                                                                                                                                                              
```

field.storage.node.lgms_banner is in optional: https://github.com/localgovdrupal/localgov_microsites_group/blob/1.x/config/optional/field.storage.node.lgms_banner.yml
field.field.node.localgov_event.lgms_banner  is in install: https://github.com/localgovdrupal/localgov_microsites_group/blob/1.x/modules/localgov_microsites_events/config/install/field.field.node.localgov_event.lgms_banner.yml

stephen-cox:
Fixes #164


ekes:
I fear putting stuff in install rather than optional will make testing much more difficult because of the quite considerable dependencies it then requires (see comment #164)

stephen-cox:
> I fear putting stuff in install rather than optional will make testing much more difficult because of the quite considerable dependencies it then requires (see comment #164)

That might be an issue, but a lot of the config in optional is required and that fact that it's not being installed in the right order is now causing issues.

ekes:
From your example in #164 can it not go in the opposite direction though? The config that's in install go to optional as well?

stephen-cox:
We certainly could move the config to optional, but is this not likely to causes issues later is something was accidentally removed and a load of config then doesn't install. If our tests don't require that thing then we're unlikely to pick it up straight away.

It seems to me that if something is required for Microsites to function as expected then it should be install. While finding all the dependencies is a pain when writing kernel tests, it wasn't too bad to add the missing dependencies here.

ekes:
> It seems to me that if something is required for Microsites to function as expected then it should be install. 

Is the banner required for Microsites to function as expected. It's something you can test in the Functional tests with the whole site bootstrapped, and it's visual, is it required at a lower level?

> it wasn't too bad to add the missing dependencies here.

Shall we try it against #160

stephen-cox:
There's code that assumes the existence of the banner fields, so if that config is in optional we'll have to change it to check whether the field has been installed.

Happy to look at those tests tomorrow

finnlewis:
Tested this and it does fix the install for me locally. 

Not sure I understand the potential problems with moving the config to install @ekes 

Is it just working out the dependencies for the tests like on https://github.com/localgovdrupal/localgov_microsites_group/pull/166/commits/96bfda46dd29ec68e3e00df1be288d7519e3815d ?

ekes:
> Not sure I understand the potential problems with moving the config to install @ekes

Have a bash at "Shall we try it against https://github.com/localgovdrupal/localgov_microsites_group/pull/160" :-P

But if this is necessary for installation, then its necessary for installation and we'll have to jump the hoops in tests. It's that or we more cleverly split up where functionality & config live.

stephen-cox:
@ekes It does look like it's just a matter of adding the extra dependencies to the `$modules` array to get kernel tests passing.

This https://github.com/localgovdrupal/localgov_microsites_group/pull/160/commits/e45235d2f1069d76030876babef2649adb81f6aa fixed the kernel tests on #160 after moving field config into install.

ekes:
> This https://github.com/localgovdrupal/localgov_microsites_group/commit/e45235d2f1069d76030876babef2649adb81f6aa fixed the kernel tests on https://github.com/localgovdrupal/localgov_microsites_group/pull/160 after moving field config into install.

:+1: Ta!

[aside: now just fighting the functional test for different reasons]

stephen-cox:
@ekes Are you happy for this to be merged now or is there more discussion to be had? Site installs are borked until we resolve this.

ekes (approved):
With the updated tests. Go for it.

PR #171, opened 2022-07-28
Add requirement on localgov_page to composer.json

### LINKED ISSUE #170: Add requirement on localgov_page to composer.json
The localgov_microsites_group module has a dependency on localgov_page, but this isn't required by Composer.

https://github.com/localgovdrupal/localgov_microsites_group/blob/c80db640d8eb06f7785593fafbfdb96abedda8c6/localgov_microsites_group.info.yml#L25

 If we want to be able to install Microsites into an existing site all the explicit dependencies should be included in the composer.json file

stephen-cox:
Fixes #170


finnlewis (approved):
Logical! Thanks @stephen-cox

PR #173, opened 2022-08-01
Permissions from https://github.com/localgovdrupal/localgov_microsites/pull/175#pullrequestreview-1032762388

ekes:
Not sure I'm happy with adding the dependency on localgov_microsites_group when it's not really needed. Should these roles maybe be in the profile? With the modules that add permissions to them silently not doing so if they don't exist?

Can't do that because we rely on them too much more? Would the hook work in the profile. It's installed before everything else I think? What's profile / starterkit -- and what's (which) module?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,6 +5,7 @@
   "license": "GPL-2.0-or-later",
   "require": {
     "drupal/core": "^9",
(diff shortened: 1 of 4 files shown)


stephen-cox (approved):
I think this is fine for now. Let's review permissions and dependencies in September.

PR #174, opened 2022-08-01
ginvite test

ekes:
Uses the trait (and hence branched from) https://github.com/localgovdrupal/localgov_microsites_group/pull/160


stephen-cox:
@ekes Re-ran the tests after the updates to Paragraphs and Page. Tests now failing with `The "group_term:localgov_event_category" plugin does not exist.` which looks like there's a missing dependency on localgov_microsites_events.

stephen-cox (approved):
Tests passing. All good to merge.

PR #177, opened 2022-08-02
Don't use two step form for nodes.

ekes:
Small one I noticed a couple of content types still using the two-step form.


stephen-cox (approved):
:+1:

PR #179, opened 2022-08-02
Require LocalGov Paragraphs 2.x-dev to fix tests

stephen-cox:


stephen-cox:
Not needed any longer since there's been a release of LocalGov Paragraphs

PR #194, opened 2022-09-01
Add role_delegation module and restrict roles controllers can assign

### LINKED ISSUE #193: Bug: News editors can access the Content Overview (and create content not in their microsite)
Ah, I realise now that we dont need to give our users "News editor" role. However it is confusing that it is available

Look at Role Assign, so that Microsite Controllers can only assign the Trusted Editor role

Document it too!

stephen-cox:
Fixes #193


msayoung (approved):
approved!

PR #197, opened 2022-09-02
Add embed paragraphs to content types

### LINKED ISSUE #196: Add embed paragraphs to content types

markconroy:
Closes #196


msayoung:
@markconroy - I've update the view to show upcoming events HOWEVER we're going to need to do something clever to alter the display of recurring events, as the teaser mode shows the same date for each.

i finally understand why localgov_events is using fields in the view!

Anyway, I think its good to merge for now and we can make another ticket for the display.

PR #200, opened 2022-09-05
Rename 'Node' and 'Group menus' site menu links

### LINKED ISSUE #198: Rename "Group menus" to "Menus"
Image

### LINKED ISSUE #199: Rename "Nodes" to "Content" (or Pages?)
Image

This got renamed at some point

stephen-cox:
Fixes #198 
Fixes #199

### DIFF
--- a/localgov_microsites_group.module
+++ b/localgov_microsites_group.module
@@ -192,6 +192,11 @@ function localgov_microsites_group_local_task_title($task_id) {
     case 'group.edit_form':
       return t('Site design');
 
+    case 'views_view:view.group_nodes.page_1':
+      return t('Content');
+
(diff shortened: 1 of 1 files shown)


msayoung (approved):
I'm happy with this, but do we mind that the URL for the Content page is /group/1/nodes 
Is that too complicated to change?

stephen-cox:
> I'm happy with this, but do we mind that the URL for the Content page is /group/1/nodes Is that too complicated to change?

Changing the path is problematic, if this is something we want it will need some investigation

willguv:
I'm OK with /group/1/nodes for now. Maybe write another card for investigation at a later date? @msayoung

PR #204, opened 2022-09-05
Added cache invalidation for menu changes

### LINKED ISSUE #94: Clear node group content cache when saving a group

### LINKED ISSUE #203: Cache clear required for menus to show up

stephen-cox:
Also removes unneeded cache invalidation for group setting changes. It looks like things have changed since #94 was implemented so the correct cache tags are invalidated when group settings are saved. I've added a test to check it's not required. 

Fixes #94
Fixes #203


msayoung (approved):
This seems to work. However I also discovered the footer menu itsn't appearing. Not connected to this MR though, will make another ticket

PR #210, opened 2022-09-07
Fix microsite alterations to group views. Add overview views.

ekes:
This modules version of the group's views for members and nodes (views.view.group_nodes/members) were originally taking precedence over groups. So there was some content on the dashboard (just content), the relation links were removed from the operations on the content page and some other changes.

At some point the install process changed to prefer the group version, so these all got lost.

This version does it differently by adding to the group views; and creating it's own version of the pages for microsite groups. This can easily be extended further if there are more microsite specific changes to group views needed. 

Fixes:
 * #214 Microsite dashboard is empty
 * #216 Remove relation from Operations
 * Also corrects node -> content references on the microsite content page.


msayoung:
@ekes - all those things look good. However the Content page has lost its local tasks and Add new content buttons

image

The other ones look fine.

msayoung:
I think we may want to show the View titles on the Dashboard to describe what we are seeing, but good for now.

ekes:
> @ekes - all those things look good. However the Content page has lost its local tasks and Add new content buttons
> 
> image
> 
> The other ones look fine.

Ah! They're probably attached to the display. I'll find them and duplicate.

ekes:
That should have covered both of them. @msayoung

msayoung:
Grand. Conflicts now, but otherwise seems good

msayoung (approved):

ekes:
Interesting hash of the merge that the github ui made of it there. Tidying.

PR #214, opened 2022-09-07
Add domain group settings plugin to override the group domain theme. #223

ekes:
Visible to controller - can admin domain groups; or users given a new permission.

### DIFF
--- a/src/Plugin/DomainGroupSettings/ThemeSettings.php
+++ b/src/Plugin/DomainGroupSettings/ThemeSettings.php
@@ -1,0 +1,198 @@
+<?php
+
+namespace Drupal\localgov_microsites_group\Plugin\DomainGroupSettings;
+
+use Drupal\Core\Access\AccessResult;
+use Drupal\Core\Extension\ThemeExtensionList;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Config\ConfigFactoryInterface;
+use Drupal\Core\Extension\ThemeHandlerInterface;
+use Drupal\Core\Language\LanguageManagerInterface;
+use Drupal\domain\DomainStorageInterface;
+use Drupal\domain_group\Plugin\DomainGroupSettingsBase;
+use Drupal\group\Entity\GroupInterface;
+use Drupal\Core\StringTranslation\StringTranslationTrait;
+use Drupal\domain\Entity\Domain;
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Drupal\Core\Session\AccountInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Drupal\domain\DomainInterface;
+use Drupal\group\Access\GroupAccessResult;
+
+/**
+ * Provides options for group domain.
+ *
+ * @DomainGroupSettings(
+ *   id = "localgov_microsites_theme_settings",
+ *   label = @Translation("Theme override"),
+ * )
+ */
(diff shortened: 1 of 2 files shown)


src/Plugin/DomainGroupSettings/ThemeSettings.php
```diff
@@ -0,0 +1,197 @@
...
+        if ($theme->getName() != $site_default) {
+          $default_options[$theme->getName()] = $theme->info['name'];
+        }
+        if ($theme->getName() != $site_admin) {
+          $admin_options[$theme->getName()] = $theme->info['name'];
+        }
+      }
+    }
+
+    $admin = $config_override ? $config_override->get('admin') : '';
```
msayoung: Can we swap these fields round, so the default theme is first?

src/Plugin/DomainGroupSettings/ThemeSettings.php
```diff
@@ -0,0 +1,197 @@
...
+      '#default_value' => $admin,
+      '#description' => $this->t("Override theme used for administration pages."),
+      '#options' => $admin_options,
+    ];
+    $default = $config_override ? $config_override->get('default') : '';
+    $form['default'] = [
+      '#type' => 'select',
+      '#title' => $this->t('Default theme'),
+      '#default_value' => $default,
+      '#description' => $this->t("Override theme used for user pages."),
```
msayoung: "Override the theme used on this microsite (default: LocalGov Microsites Base)."

src/Plugin/DomainGroupSettings/ThemeSettings.php
```diff
@@ -0,0 +1,197 @@
...
+      '#options' => $default_options,
```
msayoung, line 149: LocalGov Microsite Base doesn't appear as an option, can it please?

src/Plugin/DomainGroupSettings/ThemeSettings.php
```diff
@@ -0,0 +1,197 @@
...
+      }
+    }
+
+    $admin = $config_override ? $config_override->get('admin') : '';
+    $form['admin'] = [
+      '#type' => 'select',
+      '#title' => $this->t('Admin theme'),
+      '#default_value' => $admin,
+      '#description' => $this->t("Override theme used for administration pages."),
+      '#options' => $admin_options,
```
msayoung: Amazing, lets put Gin back in the mix (I'll raise another issue for this)

msayoung (changes requested):
Amazing @ekes. 
A couple of comments. 

Additionally, can we have a little bit of intro text to these fields. Something like: 

The default theme is LocalGov Microsites Base, which can be customised extensively through the UI on the <a>Site design</a> tab of the microsite. If you need further customisation we recommend making a child theme of LocalGov Microsites Base. (Documentation to come).

src/Plugin/DomainGroupSettings/ThemeSettings.php
ekes: The options don't show the existing default. So you don't override the selection with the value it would be anyway.

I had wondered if I made the present blank option called: 'No override (Existing theme name)' -- or something else that's clear that it's not an override.

msayoung: Yes, that would work. I can imagine someone finding the options and playing around with it and not knowing how to put it back.

ekes:
Does that cover it?

msayoung (approved):

Issue #218, opened 2022-08-30
Bug: Competing URLS (news)


msayoung:
When there are 2 microsite instances using News, the first URL defaults to /news and the second to /news-1 
The second can be updated manually, but can we override this somehow, so that it defaults to be /news as well ?

ekes:
So think this is two topics:
 * Remove the default pathauto. This probably makes sense for all content types. For example events also get /events/name /events/name-0 paths automatically. These are global, not site specific.
 * Creating domain specific automatic paths. For which there was an issue, and hopefully a reminder of why that wasn't working properly.

First one should hopefully be straightforward.
Once that is done users can set their own custom path _manually_.
Second one... ah...

ekes:
> First one should hopefully be straightforward.

Ah! Well https://github.com/localgovdrupal/localgov_microsites_group/commit/94786f616897631dc64b10b1abedee88b4164f24

ekes:
So my hunch (mentioned on the commit message https://github.com/localgovdrupal/localgov_microsites_group/commit/94786f616897631dc64b10b1abedee88b4164f24) was correct. Because we re-scan config/optional the paths there keep getting put back.

This means no matter which module is being installed the config gets put back. At the moment the best I can think of for these configurations is to delete them on _any_ module install. The only other option I can see is to delete them whenever they are created, but that seems way too brutal. Any module seems brutal, but stopping someone from manually creating them with some opaque magic... that's possibly a step too far. As mentioned in the code comment there's really no way of knowing where the configuration is coming from when it's created, sadly, and no hook or event on importing module configuration (other than the modules installed hook we're using).

@stephen-cox Any opinion? Ideas? Otherwise I'll just drop these two out of the check for which modules are being enabled.

stephen-cox:
@ekes  You mentioned at standup some modules have Pathauto config in install, so perhaps we should be consistent and do this for all modules? It will mean adding an explicit dependency on Pathauto, which may not be desirable. Other than that I can't think of a nice solution to this.

ekes:
> some modules have Pathauto config in install, so perhaps we should be consistent and do this for all modules?

I did think about this, but it would mean getting patches into existing LocalGov modules, and then thinking about it I'm not sure I can make a convincing case that it should be in config/install, rather the other way round, it's only where the breadcrumb built from the path and is a requirement that it's not optional really. In fact it is because it's actually optional we want it in install - which is mindbending.

ekes:
[...]

PR #219, opened 2022-09-09
Remove pathauto patterns that will be applied globally for microsite content

ekes:
The don't create the global pathauto path part of #218


finnlewis:
Testing locally, seems to work as expected. 

On the 2.x-dev branch we get all the path auto patterns: 

image

On the feature/pathauto-global-path branch: 

image

All seems to work!

finnlewis (approved):
Works for me testing locally. Thanks @ekes !

ekes:
Ugly, but it works.

PR #220, opened 2022-09-09
Add a file field to upload a favicon (png) and set it as the theme configuration favicon. localgov_microsites#178

ekes:
Outstanding questions on https://github.com/localgovdrupal/localgov_microsites/issues/178#issuecomment-1242329228
You should be able to test this as is; but it won't change the favicon you see until the second of those questions is covered (or you delete the lines from the html.html.twig in the localgov_base theme).

### DIFF
--- a/config/install/core.entity_form_display.group.microsite.default.yml
+++ b/config/install/core.entity_form_display.group.microsite.default.yml
@@ -6,6 +6,7 @@ dependencies:
     - field.field.group.microsite.lgms_base_line_height
     - field.field.group.microsite.lgms_base_spacing
     - field.field.group.microsite.lgms_body_font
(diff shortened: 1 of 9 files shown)


finnlewis:
This totally works for me, as long as we override the html.html.twig - should we add a PR for that in https://github.com/localgovdrupal/localgov_microsites_base ?

finnlewis (approved):
Tested locally and works well for me, as long as we override html.html.twig in https://github.com/localgovdrupal/localgov_microsites_base

finnlewis:
@markconroy if you have time and headspace to think about this and create issues or pull requests in the localgov_microsites_base, that'd be grand!

PR #221, opened 2022-09-12
Fix #217 add domain_group process to restrict search api results

### LINKED ISSUE #217: Bug: News categories and news items from other sites available
When more than one microsite instance is using News, and some Categories are in use

EG: https://site7.bathnes.localgovdrupal.org/news has 4 news items, and one is using the taxonomy term Education. If we select education we get the following page

Image

showing the correct item from the site. 

Visit: https://site2.bathnes.localgovdrupal.org/news

There is 1 news item which does not have categories and yet the categories show a count of 1

Image

Less important - as user 1 you can also see the content from other sites when you click the facet. A microsite admin does not, I haven't tested with the microsite controller.

ekes:
Requires latest domain_group to test with commit https://git.drupalcode.org/sandbox/ekes-3278349/-/commit/389d3cc0fea7525d10719d69022d7350d0475223

This adds a processor that includes the microsite group with all the indexed items, and then applies a filter by domain to them. I've only manually tested that caching is correct.

### DIFF
--- a/config/install/core.entity_form_display.group.microsite.default.yml
+++ b/config/install/core.entity_form_display.group.microsite.default.yml
@@ -6,6 +6,7 @@ dependencies:
     - field.field.group.microsite.lgms_base_line_height
     - field.field.group.microsite.lgms_base_spacing
     - field.field.group.microsite.lgms_body_font
+    - field.field.group.microsite.lgms_favicon
     - field.field.group.microsite.lgms_footer_background_colour
     - field.field.group.microsite.lgms_footer_banner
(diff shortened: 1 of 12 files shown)


finnlewis (approved):
Tested locally and works well. Noting that the favicon work is included with this pull request, but I've also review that on https://github.com/localgovdrupal/localgov_microsites_group/pull/220 and it also works and looks good. Thanks @ekes !

ekes:
Closing in favour of #224 that only contains the relevant commit.

PR #224, opened 2022-09-12
Add domain_group processor to limit results by domain. Fixes #217

### LINKED ISSUE #217: Bug: News categories and news items from other sites available

ekes:
Requires domain_group with https://git.drupalcode.org/sandbox/ekes-3278349/-/commit/389d3cc0fea7525d10719d69022d7350d0475223

### DIFF
--- a/localgov_microsites_group.install
+++ b/localgov_microsites_group.install
@@ -2,7 +2,7 @@
 
 /**
  * @file
- * LocalGov Microsites Directories install file.
+ * LocalGov Microsites Group install file.
  */
 
 use Drupal\search_api\Entity\Index;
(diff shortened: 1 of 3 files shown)


ekes:
@finnlewis This should be #217 without the favico stuff.

finnlewis (approved):
Tested again locally, works as expected. Thanks @ekes !

PR #236, opened 2022-09-16
removes default options from line height and font size fields

### LINKED ISSUE #235: Remove default options from site design fields
We have 1rem and other things set as default values in some of our site design fields. Since these then get inlined in the body element as CSS variables a custom sub-theme cannot override them, as inline takes precedence in the cascade.

markconroy:
Closes #235


msayoung (approved):

msayoung:
Was this meant to be draft @markconroy ?

PR #239, opened 2022-09-21
Update the site_name and site_slogan fields to be boolean fields.

msayoung:


msayoung:
Requires https://github.com/localgovdrupal/localgov_microsites_base/pull/114

markconroy:
Hi @msayoung 

I've updated this ever so slightly for the manage form display page to use a boolean instead of a select list. Besides that, I think we're good to go with this one.

markconroy (approved):

PR #254, opened 2022-10-04
On install select the relevant view on the localgov_embed_views parag…

msayoung:
…raph type


stephen-cox:
@msayoung this should be working now

finnlewis:
Hi @msayoung 

It installs fine and I can select the views in the embed list paragraph, but I can't get any of the views to actually display a list of blogs, events or news. Am I missing a piece of the puzzle?
Cheers,
Finn

finnlewis (approved):
Love it! Thanks @msayoung !

PR #257, opened 2022-10-05
Add the permission to delete content to editors

msayoung:


finnlewis (approved):
Tested locally and seems to work as expected, thanks @msayoung

PR #260, opened 2022-10-06
Add microsite administration link to main menu that is only shown on microsites

### LINKED ISSUE #236: Site administration menu: streamline / simplify
Feedback from user testing found the proliferation of admin menus took up valuable screen space. Also the difference in appeareance of the Microsite Local tasks was confusing (in the different themes). 

I've done a little  research and wonder if the following would be better

The toolbar_menu module allows you to place a menu as an item in the toolbar, and limit (by Drupal role) who has access to it. I've added one called "Microsite Admin" in the screenshot below.

I've used the Tools menu, but we would need to create a menu which could be referenced by this - I assume this is possible? 

Image

If we went with this we could remove the microsite local task menu from the frontend at least. 

We can also enable Local Tasks in Admin Toolbar Tools (which we're also using) which adds the local tasks to the toolbar. See the right of the screenshot above. This current shows Edit / View etc on the frontend, and all the microsite admin when in the admin theme.

@stephen-cox - you had thoughts concerns about this. What do you think? 
@willguv @markconroy @finnlewis thoughts?

stephen-cox:
Fixes localgovdrupal/localgov_microsites#236

### DIFF
--- a/localgov_microsites_group.links.menu.yml
+++ b/localgov_microsites_group.links.menu.yml
@@ -1,0 +1,7 @@
+localgov_microsites_group.microsite_admin:
+  title: 'Microsite administration'
+  description: 'Manage the current microsite'
+  menu_name: admin
+  parent: system.admin
+  weight: 2
+  route_name: localgov_microsites_group.microsite_admin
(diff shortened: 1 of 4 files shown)


finnlewis:
Just testing this locally

finnlewis (approved):
Tested this locally and it works nicely, thanks @stephen-cox !

Also means that the admin menu is no longer empty, nice! Let's get it in and see how it flies.

msayoung:
Yep, I also think it works. We could remove the tabs from the frontend too which will give us more screen space

image

Issue #255, opened 2022-10-05
Save as unpublished and other options


msayoung:
Currently our microsite Admins and editors can't save as unpublished. 

I'm planning to add https://www.drupal.org/project/override_node_options which will allow Trusted Editors to unpublish. 

Do we want to allow editors / admins to 
- set revisions
- set authored by

I can do this at the same time 

@finnlewis @willgov  

(cc @stephen-cox )

finnlewis:
I don't think the 'authored by' field is terribly useful.
Great to be able to create and review revisions.

finnlewis:
For revisions, see: https://github.com/localgovdrupal/localgov_microsites_group/issues/259

finnlewis:
Otherwise, if https://www.drupal.org/project/override_node_options  allows Editors and Admins to save as unpublished, that sounds great!

msayoung:
https://github.com/localgovdrupal/localgov_microsites_group/pull/263

PR #263, opened 2022-10-10
Allow trusted editors to create unpublished content

msayoung:
As https://github.com/localgovdrupal/localgov_microsites_group/issues/255


finnlewis:
Testing locally now

finnlewis (approved):
Tested locally and works a treat, I can now select published or unpublished when creating and editing nodes. 

Thanks @msayoung !

PR #272, opened 2022-10-14
Enable automatic generation of URLs

### LINKED ISSUE #271: Domain Path PathAuto
- [x] Create three pathauto aliases for same value, they _are_ created `string`, `string-1`, `string-2` _but_ `string-1` is also replaced as the link _to_ the content for `string-1`. At least that's my manual testing of https://www.drupal.org/project/domain_path/issues/3285213 Needs automated tests and a fix
 - [x] Pathauto checkbox is always _off_ by default, no way of changing this, you have to select it
 - [ ] We probably want to extend the automatic replacement to use group alias rather than `-1` in the first instance

_Originally posted by @ekes in https://github.com/localgovdrupal/localgov_microsites/issues/146#issuecomment-1168568432_

stephen-cox:
Fixes #271

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -42,6 +42,10 @@
         "Empty page when trying to create group node localgovdrupal/localgov_microsites_group#45": "https://www.drupal.org/files/issues/2019-01-04/2842630-20.patch",
         "Fix Deprecated function: uasort() when editing group node view. See https://www.drupal.org/node/3210808 ": "https://www.drupal.org/files/issues/2021-04-26/uasort-comparison-3210808-2.patch",
         "Node previews don't work when adding group node https://www.drupal.org/project/group/issues/2853001": "https://www.drupal.org/files/issues/2022-08-26/2853001-21.patch"
+      },
+      "drupal/domain_path": {
+          "Error manually adding an alias when Domain Path Pathauto is enabled #3315360": "https://www.drupal.org/files/issues/2022-10-14/error-manually-adding-alias-with-domain-path-pathauto-enabled-3315360-2.patch",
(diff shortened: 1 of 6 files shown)


stephen-cox:
With the changes here, Pathauto creates aliases on the controller site for all content. While this is not great, a bigger problem is all the content being available on the controller site anyway, so I suggest we discuss this on #276

finnlewis:
Not tested yet, hope to do so later today.

ekes:
[x] Creates auto aliases for domains;
[x] Remembers state of checkbox after save on initial test.
[ ] Checkbox isn't defaulted on for new posts. I'm assuming we don't need this to be configurable, we do it for microsites only in this module and just make it always on?

It seems to create the autopath for the default domain no matter the state of the checkbox in the first instance.
It uses the default domain path for the site domain as well if it's not generated a site specific one. No way to disable.

Token replacement, for 'parent post'/'entity reference' path, uses the path of the domain the item is posted on -- this is just a reason to encourage _not_ editing content on the master site, nor sharing content.

finnlewis (approved):
Yeah baby! It does seem to work on multiple microsites. 

I agree it is odd to be able to edit the content on the controller site as it needs it's domain / group context. 

Are we able to default the checkbox to be on when creating new content on the microsite?

stephen-cox:
Tests are failing because of https://github.com/localgovdrupal/localgov_microsites_base/issues/158

stephen-cox:
Tests are now failing on the f***ing media_library_edit missing schema issue. I'm just going to merge this

PR #279, opened 2022-10-18
Add documents to the paragraphs available

msayoung:


finnlewis (approved):
Tested locally, documents now available on blogs, events and news. Thanks @msayoung !

PR #281, opened 2022-10-20
Prevent access to /events if events aren't enabled

### LINKED ISSUE #275: Disable access to /events when events are turned off for a site
At the moment access to /events depends on whether there are any event nodes.

stephen-cox:
Fixes #275

### DIFF
--- a/localgov_microsites_group.module
+++ b/localgov_microsites_group.module
@@ -146,7 +146,7 @@ function localgov_microsites_group_menu_link_content_presave(MenuLinkContentInte
 
   // Clear group cache so menu links are visible.
   if ($group_id) {
-    Cache::invalidateTags(['group:' . $group_id]);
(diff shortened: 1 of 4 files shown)


stephen-cox:
Tests are failing because of https://github.com/localgovdrupal/localgov_microsites_base/issues/158

stephen-cox:
Having fixed an issue in the theme which was breaking the tests they;re now failing on the f***ing media_library_edit missing schema issue. Shouldn't effect the review of this.

finnlewis:
testing now

finnlewis (approved):
Seems to work for me locally. When I disable the events content type, I page not found at /events. With events enabled I get an empty list. 

Thanks @stephen-cox ! 

(Also noting that we are ignoring the missing schema test fails for now on the silly media_library_edit config)

Issue #252, opened 2022-09-27
Disabling News also disables events


msayoung:
Also a problem with Blogs

finnlewis:
note: the code for this might change in Group 3.x, but @ekes is surprised that this is still a big.

stephen-cox:
It looks like blogs, events and news all share some permissions to do with taxonomy terms which is why this is a problem.

I think we'll need to work out how to exclude some permissions from the logic that works out how to switch stuff off.

stephen-cox:
I reworked the logic used to disable modules in https://github.com/localgovdrupal/localgov_microsites_group/commit/45ca3dbd50535e08c06d07d98d22f493a2998e7c

Things how work when testing through the UI, but the GroupPermissionsHelperTest::testToggleModulePermissions() test now fails

finnlewis:
This will need re-writing for Group 3.x :)

finnlewis:
Closing this in favour of
https://github.com/localgovdrupal/localgov_microsites/issues/304
and

finnlewis:
This is fixed!

PR #286, opened 2022-10-24
Fix module toggle by site

stephen-cox:


finnlewis:
Testing now

stephen-cox:
> Testing now

The failing test means I'm not sure the logic changes are correct, so I'm marking this is draft as it's not ready for review yet.

finnlewis (approved):
Functionality seems good locally... I can now toggle on and off individual content types without affecting the others.

finnlewis:
@ekes maybe could you check if this works with Group 3.x?

ekes:
Note #252 related, and #296 
If all is well #305 is fixed, so now can check these.

ekes:
Tests certainly need fixing, but from a front end perspective it certainly seems to be doing the trick.

finnlewis:
In Merge Monday, @stephen-cox recalls that the tests were not testing the right thing, so we either remove the bad tests or fix the tests.

ekes:
> [Tests fixes: 3 out of 4. 1 to go.](https://github.com/localgovdrupal/localgov_microsites_group/pull/286/commits/42377a3de7f90d6ad2f85d2c6d3dc7580d5edc78)

And the 1 is breaking because it's broken, not because the test is wrong.

ekes:
Tests pass now. Suggest maybe a follow-up to see if when GroupPermission entity is saved it shouldn't automatically flush some cache (we have two drupal_flush_all_caches() in the test that might be rendered unnecessary).

finnlewis:
Coding standards now pass.

But @ekes I can't enable / disable blogs or events. 

news and directories seem to work OK.

Site 1 settings from control site: 

image

node add create page on site 1 as a microsite editor 

image

stephen-cox:
This largely looks to work fine. There is some strange caching going on. The add new content screen doesn't always get updated. After a full cache rebuild things seem to work as expected.

ekes:
Doesn't seem to be as reliable as clearing cache. There seems to be something not quite right about the way we're creating the group permissions entity. I'm prodding.

finnlewis:
This does now work for me, if I clear the cache.

Issue #298, opened 2022-11-15
[3.x] Update how 'create content' page (and others menu, directory facets)


finnlewis:
Fresh install.
Created one microsite.
Visit : https://localgov-micro.lndo.site/group/1/node/add

```
The website encountered an unexpected error. Please try again later.
Error: Call to undefined method Drupal\group\Entity\GroupRelationshipType::getDescription() in Drupal\localgov_microsites_group\EventSubscriber\AlterCreateGroupContentSubscriber->onView() (line 49 of modules/contrib/localgov_microsites_group/src/EventSubscriber/AlterCreateGroupContentSubscriber.php).

Drupal\localgov_microsites_group\EventSubscriber\AlterCreateGroupContentSubscriber->onView(Object, 'kernel.view', Object)
call_user_func(Array, Object, 'kernel.view', Object) (Line: 142)
Drupal\Component\EventDispatcher\ContainerAwareEventDispatcher->dispatch(Object, 'kernel.view') (Line: 174)
Symfony\Component\HttpKernel\HttpKernel->handleRaw(Object, 1) (Line: 81)
Symfony\Component\HttpKernel\HttpKernel->handle(Object, 1, 1) (Line: 58)
Drupal\Core\StackMiddleware\Session->handle(Object, 1, 1) (Line: 48)
Drupal\Core\StackMiddleware\KernelPreHandle->handle(Object, 1, 1) (Line: 106)
Drupal\page_cache\StackMiddleware\PageCache->pass(Object, 1, 1) (Line: 85)
Drupal\page_cache\StackMiddleware\PageCache->handle(Object, 1, 1) (Line: 48)
Drupal\Core\StackMiddleware\ReverseProxyMiddleware->handle(Object, 1, 1) (Line: 51)
Drupal\Core\StackMiddleware\NegotiationMiddleware->handle(Object, 1, 1) (Line: 23)
Stack\StackedHttpKernel->handle(Object, 1, 1) (Line: 709)
Drupal\Core\DrupalKernel->handle(Object) (Line: 19)

ekes:
/group/1/node/add

Again has a list like:-

Group node (Blog channel)
    Add existing content item of type Blog channel to the group. 

This was being altered, but breaks with 3.x the above commit [c324a5a](https://github.com/localgovdrupal/localgov_microsites_group/commit/c324a5ad1145bd2e5e3550c919247961c761a39f) stops it breaking but doesn't fix the change.

[...]

PR #318, opened 2022-11-20
Override presentation of Bundle name and Description for Group Plugins generally. #298

ekes:
Should make all the add content pages look ok (as much to #298 as there is info).


msayoung (approved):
Stunning, except I'm seeing the wrong list of content types. 
However I think that's fixed elsewhere

PR #323, opened 2022-11-21
Update trusted editor permissions to allow revisions edit, revert, de…

msayoung:
…lete


ekes (approved):
If they should have access that makes sense to me.

PR #325, opened 2022-11-21
Update views for group relationships (3.x) #324

### LINKED ISSUE #324: Some Embddedable views not installing
The lgms_embed_view_events and lgms_embed_view_news views are not installing. 

lgms_embed_view_blogs is installing fine. 

_might_ be the     - group.relationship_type.group_relationship_type_5a3c3717 dependency ?

ekes:
This certainly fixes #324 question is if the output of the view is correct. It now uses the relationship rather than a filter.


finnlewis:
I will test the output of the events and the blogs views

finnlewis (approved):
tested locally and the embedded views work fine. Thanks @ekes  and @msayoung ! :)

PR #327, opened 2022-11-21
Fix taxonomy listing view so that it does not require the permission …

finnlewis:
…to view the group releationship entity to view terms.

This should mean that the user no longer needs permissions to view group relationships for the terms in the group, just to see the listed terms.

Happy to merge and test, but figured a pull request helps to share.

### DIFF
--- a/modules/localgov_microsites_group_term_ui/config/install/views.view.lgms_group_taxonomy_terms.yml
+++ b/modules/localgov_microsites_group_term_ui/config/install/views.view.lgms_group_taxonomy_terms.yml
@@ -1,20 +1,22 @@
 langcode: en
 status: true
 dependencies:
+  config:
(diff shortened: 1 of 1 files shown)


finnlewis:
I got the tests running... which reveals a few issues to look at. 

@ekes how best to deal with the test fails? 

- Schema issues like views.view.group_terms:display.default.display_options.relationships.gc__taxonomy_term.group_relationship_plugins missing schema
- Drupal\Tests\localgov_microsites_events\Functional\MicrositeEventViewAccesTest::testMicrositeEventsViewAccess
ParseError: syntax error, unexpected 'protected' (T_PROTECTED), expecting variable (T_VARIABLE)

Shall we create separate issues for each or are some quick ?

PR #336, opened 2022-11-22
Remove responsive preview

msayoung:


finnlewis (approved):
Works for me, thanks @msayoung

PR #343, opened 2022-11-24
Add a page bottom group for each content type on install and add fiel…

msayoung:
…ds to it


msayoung:
To test what this is doing:

- creates a field group on the content types called Page Bottom, puts some fields within it. this is so some fields always appear at the end of the content.

https://github.com/localgovdrupal/localgov_microsites_base/issues/174
https://github.com/localgovdrupal/localgov_microsites_base/issues/172

@finnlewis can you see how to fix the remaining coding standard error ?
Also not sure about the failing tests

finnlewis (approved):
Works for me locally, thanks @msayoung

PR #361, opened 2022-12-05
Ability to accept invites.

ekes:
Avoid needing to see the invites to accept them (there's no permission to do so); and one less step that's not required anyway.

### DIFF
--- a/config/optional/group.relationship_type.microsite-group_invitation.yml
+++ b/config/optional/group.relationship_type.microsite-group_invitation.yml
@@ -7,15 +7,15 @@ dependencies:
     - ginvite
     - user
 id: microsite-group_invitation
-label: 'Microsite: Group Invitation'
-description: 'Creates invitations to microsite.'
 group_type: microsite
 content_plugin: group_invitation
 plugin_config:
   group_cardinality: 0
   entity_cardinality: 0
-  use_creation_wizard: false
   unblock_invitees: true
+  invitation_bypass_form: true
+  invitation_expire: null
+  invitation_expire_keep: false
   invitation_subject: 'You have a pending microsite invitation'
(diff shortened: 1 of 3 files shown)


msayoung:
Hmm - still not working. 
Also having a conflict now that we've disabled Anons from creating accounts.

ekes:
We do want this, but yes it's not going to do all we need. 
Replacing the view is the best option that we can package now.

We do need to have people create accounts though... unless we investigate also wiring into something that generates the account and sends a resend password link.

ekes:
> We do need to have people create accounts though... unless we investigate also wiring into something that generates the account and sends a resend password link.

The configuration was that by default new account requests went to moderation; but those that had invites got accepted directly with a one-time login link. So I think it does want to go back to that, even if it means that you might end up with a moderation queue of spam attempts.

ekes:
Pushed commit should make it possible to get to page of invites and click on accept.

The UX is rubbish still though. The invite isn't descriptive, and on accept it just returns to the my invites page.
We are however now overriding the view so we could do it further. Redirecting to the group accepted would be one. The fact that you can't see what you are accepting is another, but probably more complicated.

msayoung:
@ekes seems to work fine. As you say not a great UX. 
There's a coding standard that needs sorting. 

Have another PR here to require admin approval for vistor registrations: https://github.com/localgovdrupal/localgov_microsites/pull/331

msayoung (approved):

finnlewis:
> @ekes seems to work fine. As you say not a great UX. There's a coding standard that needs sorting.
> 
> Have another PR here to require admin approval for vistor registrations: [localgovdrupal/localgov_microsites#331](https://github.com/localgovdrupal/localgov_microsites/pull/331)

I merged that one @msayoung

finnlewis:
I've changed the function name for that coding standards fail.

PR #362, opened 2022-12-05
Allow anonymous users to see taxonomy terms

msayoung:


ekes (commented):
Makes sense that it's viewable by anon. Should it not also be for OUTSIDER too? Less important as it's a bit of an edge case that we're trying to avoid ever happening, but for when it does?

ekes:
Looks like the test needs the expected permissions array updating too.

finnlewis (approved):
Looks sensible, thanks @msayoung

PR #365, opened 2022-12-06
Update the anon and outsider permissions to allow them to view group …

msayoung:
…media entities


finnlewis (approved):
LGTM! Thanks @msayoung

PR #367, opened 2022-12-06
Pin versions of external dependencies on dev or alpha tags.

finnlewis:


finnlewis:
Flagging this up to @ekes , @stephen-cox and @msayoung as it might be of interest. Pinned dependencies to specific tags or commits.

composer.json
```diff
@@ -7,18 +7,18 @@
...
     "drupal/domain_path": "1.x-dev#ba7b37af",
     "drupal/field_formatter_class": "^1.5",
-    "drupal/ginvite": "^4.0@alpha",
-    "drupal/group": "3.0.x-dev",
-    "drupal/group_content_menu": "3.x-dev",
-    "drupal/group_permissions": "2.0.x-dev",
-    "drupal/group_term": "dev-3316529-group-3.x-compatibility",
-    "drupal/groupmedia": "^4.0@alpha",
+    "drupal/ginvite": "4.0.0-alpha1",
+    "drupal/group": "3.0.0-beta4",
```
ekes: I'm guessing this should be stable enough we can ^3.0.0-beta4 it.

ekes (approved):
Rest looks correct don't want any of the alpha modules automatically updating, and then the rest are pinned to commits.

PR #368, opened 2022-12-06
Only turn domain path auto on by default for the active domain 

### LINKED ISSUE #338: Domain path aliases no saving
I have two /services on two different microsites 

The second one defaults to /services-0 but I can't override it. 

Image

Image

stephen-cox:
Fixes #338


finnlewis (approved):
Works for me, thanks @stephen-cox !

PR #373, opened 2022-12-12
2.0.0-beta2

ekes:
* Switch to release of group_term with the correct group dependency.


ekes:
The localgov_microsites_project wants releasing first to stop overriding the repository that group_term is coming from.

PR #375, opened 2022-12-16
Fixed typo in service name

### LINKED ISSUE #282: Typo in localgov_microsites_group.redirect_subscriver
Change from
```localgov_microsites_group.redirect_subscriver:```
to 
```localgov_microsites_group.redirect_subscriber:```

Then fix up anywhere it is used.

stephen-cox:
Fixes #282


ekes (approved):
grep'ing the code this is the only use of the service string, it is an event_subscriber so that makes sens.

PR #380, opened 2023-01-09
Update GitHub Actions to Composer install after requiring everything

stephen-cox:


stephen-cox (commented):
Update GitHub authentication config

finnlewis:
@stephen-cox do we still want this?

Issue #376, opened 2022-12-16
Tests failing: Undefined array key "gc__taxonomy_term"


stephen-cox:
Tests are failing with `Exception when installing config for module group, message was: Undefined array key "gc__taxonomy_term"`

Looks to be a result of merging #372

finnlewis:
Aha - here's a reference to the `gc__taxonomy_term` error.

@ekes do you know of any differences that might have introduced this with https://github.com/localgovdrupal/localgov_microsites_group/pull/372 ?

ekes:
> Aha - here's a reference to the `gc__taxonomy_term` error.
> 
> @ekes do you know of any differences that might have introduced this with #372 ?

#372 does:-

https://github.com/localgovdrupal/localgov_microsites_group/pull/372/files
```
-    "drupal/group_term": "dev-3316529-group-3.x-compatibility#149bf98c",
+    "drupal/group_term": "4.0.0-alpha2",
```
So checking out git@git.drupal.org:issue/group_term-3316529.git @ 149bf98c, and adding the remote for the main module, and then diff'ing. Ohh! There's lots of stuffs. But it's mainly exported views, and text changes.

The exported views.view.group_terms does mention (in both versions `gc__taxonomy_term`)
[group-term-3316529at149bf98.diff.txt](https://github.com/localgovdrupal/localgov_microsites_group/files/10447587/group-term-3316529at149bf98.diff.txt)

ekes:
https://github.com/localgovdrupal/localgov_microsites_group/blob/2.x/composer.json#L14 should be alpha3 by now btw

finnlewis:
I tested yesterday with the original branch from #372 (https://git.drupalcode.org/issue/group_term-3316529/-/tree/3316529-group-3.x-compatibility) and I got the same failures locally, so it does not seem to be directly due to the switch.

finnlewis:
I've posted an issue over at https://www.drupal.org/project/group_term/issues/3334882 

I've tried removing everything from the test, and just adding the dependency on localgov_microsites_events is enough to trigger the error. 

Some bad config in there perhaps? Or a problem with the group_term module.

```
/**
   * {@inheritdoc}
   */
  protected static $modules = [
    'localgov_microsites_events',
  ];
```

finnlewis:
[...]

PR #383, opened 2023-01-23
Retrieve non-overridden settings to find default.

### LINKED ISSUE #382: Saving "Site settings" is resetting theme override
## How to reproduce
0. Create a microsite as a user with the microsite-controller role.
0.5.  Login to the new microsite.
1. Override the microsite theme from the "Site settings > Theme override" tab.  Note down the newly selected microsite theme name.
2. Edit the site name or site slogan from the "Site settings > Site settings" tab.  Save the form.
3. Reopen the "Site settings > Theme override" tab and compare the selected theme with the one noted down in step 1.  These will defer.

Copying @clearspotmedia @ekes @finnlewis

ekes:
Otherwise it would seem the override is the default, at which point settings get removed as it's assumed to be ''. 

Fixes #382

### DIFF
--- a/src/Plugin/DomainGroupSettings/ThemeSettings.php
+++ b/src/Plugin/DomainGroupSettings/ThemeSettings.php
@@ -103,7 +103,8 @@ class ThemeSettings extends DomainGroupSettingsBase implements ContainerFactoryP
       }
     }
 
-    $site_config = $this->configFactory->get('system.theme');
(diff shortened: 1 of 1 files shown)


finnlewis (approved):
Works well for me locally now, thanks @ekes 

The errors in the tests are not related to this pull request (but are really annoying!) and are being discussed on this issue: https://github.com/localgovdrupal/localgov_microsites_group/issues/376

PR #386, opened 2023-02-13
Return to group edit page when saving a group

### LINKED ISSUE #359: Saving site design sends you to dashboard tab
Whenever you save a site design, it goes back to the dashboard. 

When testing designs this is frustrating. 

Can the save be set to reload where it is, rather than send you to dashboard?

stephen-cox:
Fixes localgovdrupal/localgov_microsites#359

### DIFF
--- a/localgov_microsites_group.module
+++ b/localgov_microsites_group.module
@@ -255,6 +255,13 @@ function localgov_microsites_group_form_media_library_add_form_oembed_alter(&$fo
 }
 
 /**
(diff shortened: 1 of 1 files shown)


ekes (approved):
Notes:
 - was concerned about it effecting other situations, but it's just the edit form
 - still redirects to the microsites list if you hit edit from there (it uses a `?destination=/admin/microsites` which overrides this

PR #389, opened 2023-03-16
initial commit for hiding breadcrumbs on homepage

### LINKED ISSUE #388: Option to hide breadcrumbs on homepage
Tandem issue with https://github.com/localgovdrupal/localgov_microsites_base/issues/207

markconroy:
Closes #388

### DIFF
--- a/config/install/core.entity_form_display.group.microsite.default.yml
+++ b/config/install/core.entity_form_display.group.microsite.default.yml
@@ -59,6 +59,7 @@ dependencies:
     - field.field.group.microsite.lgms_heading_6_line_height
     - field.field.group.microsite.lgms_heading_font
     - field.field.group.microsite.lgms_heading_font_weight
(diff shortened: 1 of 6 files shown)


markconroy:
@ekes and/or @stephen-cox  we might need to add an update hook to add this to existing sites.

finnlewis:
@keelanfh asks if we should default the checkbox to be checked to 'hide the breadcrumb on the homepage'. 

@Adnan-cds thinks yes, default to checked.

finnlewis:
@ekes am I right in thinking we're taking a simpler approach to hiding breadcrumbs? 

@markconroy see https://github.com/localgovdrupal/localgov_microsites_base/pull/214

I think if we go with that pull request, this one might be redundant, no?

ekes:
> I think if we go with that pull request, this one might be redundant, no?

I see a +1 on that one so yes if Mark's happy.

PR #391, opened 2023-03-21
Added dummy menu_link_reference module and update hook to uninstall it

stephen-cox:
Fixes #390

PR #392, opened 2023-03-21
Use group_term alpha5

### LINKED ISSUE #376: Tests failing: Undefined array key "gc__taxonomy_term"

stephen-cox:
Fixes #376

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -46,8 +46,11 @@
       "drupal/groupmedia": {
           "Update addPage in groupmediato work with Group 3.x and suppress errors #3325129": "https://git.drupalcode.org/issue/groupmedia-3325129/-/commit/194f7bfa1d51f9a34a00ab4323f5fdf8c82ff48b.patch"
       },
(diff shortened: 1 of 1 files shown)


stephen-cox:
Fixes the 'Undefined array key "gc__taxonomy_term"' failures, but tests are still failing. New failures should be fixed with #391

finnlewis:
Nice work @stephen-cox 

We now just have this other error in the phpunit tests, which I think is related to this: 
https://www.drupal.org/project/group_content_menu/issues/3333968

stephen-cox:
> We now just have this other error in the phpunit tests, which I think is related to this: https://www.drupal.org/project/group_content_menu/issues/3333968

Should be fixed with #391

finnlewis (approved):
Very cool!

PR #401, opened 2023-04-24
Switch to updating the view pre-insert. Possible fix for #400

ekes:
@stephen-cox only done a cursory test, but it did override the view on install.

Does it fix the issues?

### DIFF
--- a/localgov_microsites_group.install
+++ b/localgov_microsites_group.install
@@ -5,6 +5,7 @@
  * LocalGov Microsites Group install file.
  */
 
+use Drupal\Core\Serialization\Yaml;
 use Drupal\search_api\Entity\Index;
 
 /**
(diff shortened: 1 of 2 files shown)


stephen-cox (approved):
Thanks @ekes. This fixes the views on a site install.

stephen-cox:
This fixes the issue, but will require running a site install to get the updated views. Just checking in Slack that this is okay. If it is let's merge and release ASAP

Boosmith:
Hi @stephen-cox If we do reinstall now, would that be it so we can then not worry about having to reinstall again? If so, that would be OK I think, as we haven't got work that can't be lost at the moment.

stephen-cox:
We are trying to ensure that there isn't a need to reinstall as microsites are in beta so there should be an upgrade path.  We should really write an upgrade hook, will see if I can get it done tomorrow.

stephen-cox:
@ekes Just pushed an update hook to apply the missing views config. Can you have a look and let me know if it looks okay?

@Boosmith Are you able to test this? If you checkout this branch and run the database updates it should fix the content, member and invitation views.

Boosmith:
Hi @stephen-cox I tried this locally with the database from our "live" site and it seemed to do the trick

ekes:
Reading the update it looks good too. I might also have time to test it later today. But as it's more pressing I'd also say good to merge now.

stephen-cox:
Thanks both - I'll merge and create a release now

Issue #396, opened 2023-03-30
Remaining test failures


stephen-cox:
Now we've resolved a bunch of issues that were causing tests to fail (#376, #390, #393), it's exposed some further test failures:

```
$ phpunit web/modules/contrib/localgov_microsites_group/
PHPUnit 9.6.6 by Sebastian Bergmann and contributors.

Testing /app/web/modules/contrib/localgov_microsites_group
.......EE...........FF....                                        26 / 26 (100%)

Time: 28:00.907, Memory: 16.00 MB

There were 2 errors:

1) Drupal\Tests\localgov_microsites_group_term_ui\Functional\ManageGroupTermsTest::testGroupTermUi
Behat\Mink\Exception\ElementNotFoundException: Button with id|name|label|value "Save and go to list" not found.

/app/web/core/tests/Drupal/Tests/WebAssert.php:148
/app/web/core/tests/Drupal/Tests/UiHelperTrait.php:77
/app/web/modules/contrib/localgov_microsites_group/modules/localgov_microsites_group_term_ui/tests/src/Functional/ManageGroupTermsTest.php:86
/app/vendor/phpunit/phpunit/src/Framework/TestResult.php:728

2) Drupal\Tests\localgov_microsites_group_term_ui\Functional\ManageGroupTermsTest::testGroupTermPermissions
Behat\Mink\Exception\ExpectationException: Current response status code is 500, but 403 expected.

/app/vendor/behat/mink/src/WebAssert.php:794
/app/vendor/behat/mink/src/WebAssert.php:130
/app/web/core/tests/Drupal/Tests/WebAssert.php:818
/app/web/modules/contrib/localgov_microsites_group/modules/localgov_microsites_group_term_ui/tests/src/Functional/ManageGroupTermsTest.php:124
/app/vendor/phpunit/phpunit/src/Framework/TestResult.php:728

--

There were 2 failures:

1) Drupal\Tests\localgov_microsites_group\Kernel\GroupPermissionsHelperTest::testGetPermissions
Failed asserting that true is false.

/app/vendor/phpunit/phpunit/src/Framework/Constraint/Constraint.php:122
/app/vendor/phpunit/phpunit/src/Framework/Constraint/Constraint.php:55
/app/web/modules/contrib/localgov_microsites_group/tests/src/Kernel/GroupPermissionsHelperTest.php:141
/app/vendor/phpunit/phpunit/src/Framework/TestResult.php:728

[...]

PR #404, opened 2023-05-31
Require Domain Path 1.2 and remove included patch

### LINKED ISSUE #403: Use Domain Path 1.2 and remove failing patch
This is also required and patched in the profile. We need to match versions.

There's also a failing patch for the module that needs removing.

stephen-cox:
Fixes #403

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,7 +5,7 @@
   "license": "GPL-2.0-or-later",
   "require": {
     "drupal/core": "^9",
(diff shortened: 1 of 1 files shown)


stephen-cox:
The test failures here are known about: https://github.com/localgovdrupal/localgov_microsites_group/issues/396

ekes (approved):

PR #405, opened 2023-06-05
Release 2.0.0-beta10

stephen-cox:


finnlewis:
@stephen-cox do we want to address these test failures before release ?

finnlewis:
note: the test failures on this are known and discussed on https://github.com/localgovdrupal/localgov_microsites_group/issues/396

We will follow up there to fix failing tests.

PR #407, opened 2023-07-05
Feature: Controller site indicator

### LINKED ISSUE #372: Add environment indicator module
@Adnan-cds and Design Dave suggested that we could include the https://www.drupal.org/project/environment_indicator module. 

This could help avoid confusion between the control site and the microsite.

Adnan-cds:
A short warning message for the admin toolbar.  Appears on control sites only. Warns against editing content.

Closes https://github.com/localgovdrupal/localgov_microsites/issues/372


stephen-cox (approved):
Nice one @Adnan-cds - this is a nice simple way to add a message to the control site.

stephen-cox:
Tests will get a bit further (but still fail) with #410

PR #417, opened 2023-09-06
Add 'access microsites overview' permission to control access to the microsites view

### LINKED ISSUE #416: Microsites view uses nonexistent permission 'bypass group access'
The microsites overview view (/admin/structure/views/view/localgov_microsites_overview) restricts access to the 'bypass group access' permission, but this permission no longer exists.

image

~As this view is supposed to only be accessible to Microsite Controllers I suggest changing the access restrictions to this role~

I suggest adding a new 'access microsites overview' permission and use this to control access.

stephen-cox:
Fixes #416


stephen-cox:
Test failures are the expected ones so this is ready for review.

stephen-cox:
This change means user 1 gets an access denied message when clicking on the microsites menu link, which is not desirable. Might be better to go back to permissions.

stephen-cox:
I've updated the PR to add a new 'access microsites overview' permission and use this to control access to the view.

Adnan-cds:
Code changes look good.  Has no effect on my local microsites instance after code update.  Nothing breaks when I make the permission changes manually.

If these config changes are necessary for existing sites then a programmatic update of the `user.role.microsites_controller` and `views.view.localgov_microsites_overview` configs should help.

finnlewis:
@Adnan-cds argues that we need an update hook to apply the new permissions. 

This is a change to controller role, which means any controller will want this new permission to be able to continue to manage the sites.

stephen-cox:
> @Adnan-cds argues that we need an update hook to apply the new permissions.

That's fine. We'll need an update hook that creates the new permission, assigns this to controllers and updates the view to use the new permission. Probably want a test to verify this as well.

finnlewis:
@stephen-cox are you planning to add some updates hooks here or hoping for someone else to help here?

stephen-cox:
@finnlewis @Adnan-cds I have added an update hook to apply this change. I've done this on the 2.x branch as this will help people who haven't upgraded their microsites install to 3.x.

With the upgrade the none existent 'bypass group access' permission is removed, breaking access to the view. I'll also port this to the 3.x branch.

If possible, it would be good to have this reviewed and released on Monday - assuming you're happy with it, of course.

Adnan-cds (approved):
Tested the fix in our site and it works fine.  Thanks a lot :)

PR #420, opened 2023-09-08
Release 3.0.0-alpha1

stephen-cox:


stephen-cox:
Just need to fix the test and then it's ready for review.

stephen-cox:
Marking this as ready for review now. It's not quite ready to be merged (we need a release of LocalGov Forms first), but there are now only 2 failing tests and they effect the 2.x branch as well https://github.com/localgovdrupal/localgov_microsites_group/issues/396.

finnlewis (approved):
I think we're ready to merge this and tag the alpha1.

It works locally and the two remaining phpunit test fails we know are on both 2.x and 3.x and porbably to do with upstream group_permissions issues, unless anyone can rewrite the tests to functional browser tests. (see https://github.com/localgovdrupal/localgov_microsites_group/issues/396#issuecomment-1739399147 )

So let's do it and tag the alpha.

finnlewis:
@ekes what do you say?

finnlewis:
Gonna merge this, the only two failing tests we might want to rewrite as functional tests, see https://github.com/localgovdrupal/localgov_microsites_group/issues/427

PR #430, opened 2023-10-27
Fix access to the microsites overview view

stephen-cox:
The 'bypass group access' permission was removed before the upgrade to 3.x and disappears entirely in the 3.x branch. Access to the microsites overview view is broken when upgrading until access in the view is fixed. This MR will help anyone upgrading as it will fix the view access settings.

There's a PR to add an identical upgrade hook to the 2.x branch.

Part of #416


ekes (approved):
Checked view config matches.

PR #438, opened 2024-01-16
Feature/group sites

ekes:
Switches to use group_sites in place of domain_group; and removes dependency on group_permissions switching to controlling access on routes per domain for creating content.

PR #439, opened 2024-02-13
Feature/update with 2.x

ekes:
From the errors being reported by people updating from 3.x to 4.x (proposed) on https://github.com/localgovdrupal/localgov_microsites_group/pull/438 I noticed that 3.x was missing commits on 2.x (including the version of the update hook that was being reported in the update).

This merges 2.x including the new version of the update hook. We then _need_ to tag this as a new release. 3.0.0-alpha2 or people are unlikely to get the update hook so will still get the errors when testing upgrades.


stephen-cox (approved):
Fixes issues with the upgrade to proposed 4.x :+1: 

The errors here already exist. See the PR for a 3.x-alpha2 release (https://github.com/localgovdrupal/localgov_microsites_group/pull/431) so I reckon we merge and look at those separately.

PR #442, opened 2024-02-23
441/add media via paragraphs

### LINKED ISSUE #441: Can't add media via paragraphs fields
If I click on the "media" tab from "Microsites administration" I can add media to my group and use it in paragraphs (such as "Image" paragraph). But if I try to add a new image via the image paragraph type in the page components field, it doesn't get added to the group.

markconroy:
Closes #441

### DIFF
--- a/config/install/group.relationship_type.microsite-75b95924cbf69473fe8a92.yml
+++ b/config/install/group.relationship_type.microsite-75b95924cbf69473fe8a92.yml
@@ -12,5 +12,6 @@ group_type: microsite
 content_plugin: 'group_media:remote_video'
 plugin_config:
   group_cardinality: 0
+  tracking_enabled: 1
   entity_cardinality: 1
   use_creation_wizard: false
--- a/config/install/group.relationship_type.microsite-group_media-document.yml
+++ b/config/install/group.relationship_type.microsite-group_media-document.yml
@@ -12,5 +12,6 @@ group_type: microsite
 content_plugin: 'group_media:document'
 plugin_config:
   group_cardinality: 0
+  tracking_enabled: 1
   entity_cardinality: 1
   use_creation_wizard: false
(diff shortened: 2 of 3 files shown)


ekes (commented):
Looking at the code for groupmedia how were we using it without this?

Edit: Answer [we were calling it](https://github.com/localgovdrupal/localgov_microsites_group/blob/9234f000f330bde49a5df155c1b116696b27806f/localgov_microsites_group.module#L303) in [that callback](https://github.com/localgovdrupal/localgov_microsites_group/issues/441#issuecomment-1959879357) which wasn't getting the Group ID any more. Why was this preferable to enabling groupmedia standard behaviour? Cases where the standard behaviour wasn't calculating the group context correctly would be my first guess.
If we do switch though, we should remove the submit handler.

finnlewis:
Sorry, not had a chance yet, but will test this soon.

ekes:
In one of those how did that work. The [call we are making ](https://github.com/localgovdrupal/localgov_microsites_group/blob/b85a74bdae435aaa9b05022973becf2ba62582fa/localgov_microsites_group.module#L305) relies on [a plugin being returned](https://git.drupalcode.org/project/groupmedia/-/blob/8.x-2.x/src/AttachMediaToGroup.php?ref_type=heads#L158) that <[for some time](https://git.drupalcode.org/project/groupmedia/-/commit/dd92ffbadad1d73a1aed786744c50a36bada479e#592a38ba4c92c78e3680d2c110b9670d20d90d41_289_312) requires that to be checked. So our submit handler needs it. In addition to any groupmedia attempts to attach the correct group.

ekes:
And this is bring stuff in from the 2.x / 3.x branches we can't use. Let me rebase it back and make a new (additional PR).

ekes:
So #443 for 4.x and this one for 3.x

stephen-cox (approved):
:+1:

PR #446, opened 2024-03-11
allows filtering by node title on content listing page

### LINKED ISSUE #445: Allow searching for content by node title and author on the content listing page
At the moment you can only filter by content type and publishing status.

Update: let's also add in filtering by author to this issue. https://github.com/localgovdrupal/localgov_microsites/issues/448

markconroy:
Closes #445 

Screenshot 2024-03-11 at 16 06 26


stephen-cox:
@markconroy I'm having a look at this, but it's based from and going into the 3.x branch.

The default branch has recently become the 4.x branch and we don't want people installing the 3.x branch any more. Can you update this to use the 4.x branch?

markconroy:
@stephen-cox  that's updated to 4.x now.

stephen-cox:
I've given this a whirl and it doesn't look like the override config is getting applied, although I haven't had any time to check what the issue is. Will need to investigate.

ekes:
> I've given this a whirl and it doesn't look like the override config is getting applied, although I haven't had any time to check what the issue is. Will need to investigate.

At the moment we're just importing the new displays:

```
  if ($view->id() == 'group_nodes') {
    $group_node_view = Yaml::decode(file_get_contents($module_path . '/config/overrides/views.view.group_nodes.yml'));
    $displays = $view->get('display');
    $displays['microsite_page'] = $group_node_view['display']['microsite_page'];
    $displays['microsite_dashboard_embed'] = $group_node_view['display']['microsite_dashboard_embed'];
    $view->set('display', $displays);
  }
```
https://github.com/localgovdrupal/localgov_microsites_group/blob/9234f000f330bde49a5df155c1b116696b27806f/localgov_microsites_group.module#L55C1-L61C4

So we should decide if we're going to replace the whole view, or continue to patch into it.

markconroy:
I've updated this PR to include the author field _and_ the title field, but as @stephen-cox says above, when we run a fresh install of the site these changes are not getting picked up.

markconroy:
On further investigation, if you try to manually import this as a single config item via the UI, you will get the following error:

``` An entity with this machine name already exists but the import did not specify a UUID. ```

I wonder is that also what is happening when trying to do an install?

ekes:
> At the moment we're just importing the new displays

The code snippet above picks _part_ of the view and applies it to the existing view. It's _not_ importing the whole view, it's using the exported view to get configuration to patch into the configuration in the site. To do more that code needs to be told which other bits of the view to override. Or be rewritten to replace all bits. But what you see is that you can't just import the whole view, because it already is supplied elsewhere, hence the code to take bits and override them. Perhaps recipes does this better?

ekes:
The changes to the view were all in the default view display. Because it's easy I've just added that to be overridden too. I think this includes everything that got changed.

My notes from the commit message above:

[Also override the defalt view display.](https://github.com/localgovdrupal/localgov_microsites_group/pull/446/commits/fbff7cdc670ec775d49ea6bb8b7a16ee83f39914)
[fbff7cd](https://github.com/localgovdrupal/localgov_microsites_group/pull/446/commits/fbff7cdc670ec775d49ea6bb8b7a16ee83f39914)

Question: Is this wise?

If the page view in the group module changes upstream, with assumptions
about the default this could break.

Seen the other way it stops the deault breaking our view.

Alternative? Override the other parts of the view that are being added
in this MR in our displays, as has been done so far.

ekes:
Failed tests on this are https://www.drupal.org/project/viewsreference/issues/3445081#comment-15626222 I've not got the head space at the moment to do more than ask a question on that issue. Should we just ignore that schema?

More important question is if we should be overriding the default display of the groups view (above).

finnlewis:
@stephen-cox thinks this is probably fine, but take another look.

stephen-cox (approved):
So we're now replacing all the displays in the `views.view.group_nodes` view, which is originally supplied by the `gnode` module.

There are safer ways to do this – we could either create our own view or just add an extra display to the view – but as it's very unlikely any would want to use this view outside of microsites I think this is fine.

PR #453, opened 2024-04-03
Fix coding standards.

finnlewis:
Not quite there yet, but creating merge request for peer review.

PR #454, opened 2024-04-11
Fix/4.x/tests

ekes:
A companion piece to #453 

I'll make another MR with both.

And test against the container too.


ekes:
Going to close this in favour of the combined #455

PR #455, opened 2024-04-11
Fix/4.x/tests and coding standards

ekes:
Not tested locally yet. It's soooo slow. So I'll get github to have a go too.


finnlewis (approved):
Nice work @ekes !

Let's merge this baby!

PR #456, opened 2024-04-12
4.0.0-alpha3 release

finnlewis:
I'd like to do a release now that we've fixed the coding standards and phpunit tests.


ekes:
Can we add #457 ? Well, check and add.

finnlewis:
Yes! Thank you @ekes ! We've got all green lights!

Merging.

PR #457, opened 2024-04-12
Simplify getting the group from context.

ekes:
Rather than getting id and reloading group often.

Just as I was starting to use the method for something else. Thought I'd directly simplify this.

As yet untested. I've just written the code. So needs tests to run, and another set of eyes on it.


finnlewis (approved):
Reading through the code, it all looks logical and sensible. And the tests pass! Nice one @ekes !

PR #459, opened 2024-04-18
Rationalise tests to use group ids 1, 2, 3, 4, 5 and add test for admin access

finnlewis:
@ekes I was getting confused in the tests where the groups[$key] was off by one to the $group->id(), so tweaked them to align.

Is this OK? 

Could you pass your eyes over it and also tell me if the [GroupAdminAccessTest](https://github.com/localgovdrupal/localgov_microsites_group/pull/459/files#diff-70abab5ecd8e0766c300e5fd708d10e6b2fd8d9f00a9f8735261fed472e5680a) is doing what I think it is doing?

Thank you!


ekes:
Tests pass so we can merge.

PR #461, opened 2024-04-20
Add and enable antibot for some default spam protection out of the box.

finnlewis:


ekes:
Well the failing tests suggest it was correctly installed, even though I don't see the config for enabling it on log in, it's blocking the test bot from logging in.

ekes:
What's the motivation for this one?

If either project would want some out of the box protection I'd assume it'd be the localgov project. Setting up microsites is much more involved, so if you can't add your preferred anti-spam to it I'd wonder if you're able to configure the server?

ekes:
In addition to above. If it does want to get added it needs some way to switch it off for all Functional tests. The failures are because the test bot can't log in.

markconroy:
I'm with @ekes 

If this is needed it would probably also be needed on LGD CMS sites, not just microsites (which probably have fewer forms and some may have no forms at all).

Let's run it through our "should we add this module" filters to see what happens.

markconroy:
~Moving this to the LocalGov project.~ hmmm, looks like you can't move a PR, obvs.

finnlewis:
> I'm with @ekes
> 
> If this is needed it would probably also be needed on LGD CMS sites, not just microsites (which probably have fewer forms and some may have no forms at all).
> 
> Let's run it through our "should we add this module" filters to see what happens.

My thinking was: 

1. We include webforms in a default localgov_microsites install, with a default contact form.
2. We don't include webforms by default on localgov. 
3. it makes sense to ship with some default spam protection to encourage best practice. 

I think we should also do the same with localgov

PR #469, opened 2024-05-26
Tests for enabling and disabling modules and access to their content types

### LINKED ISSUE #464: Add phpunit tests to cover enabling and disabling content types for a given microsite.
To get to a stable release, @ekes suggested that we should add phpunit tests to confirm that the content type options are being respected.

As a microsites controller, 

I want to enable or disable content types for a given microsite

So that I can restrict the content types available to a microsite content designer.

As a 'trusted editor'

I want to only be able; to create see the content types that are enabled.

So that I don't create illegal content types 

So we want some phpunit tests as follows.

1. Create a microsite (micsrosite 1) as a user with the 'microsites controller' role.
2. Disable all optional content types on the microsite (group/1/domain-settings)
3. Create a user 'editor 1' and assign that user to microsite 1. 
4. Log into microsite 1 as editor 1
5. Confirm I cannot create News, Webform, Events, Directories, Blogs
8. 8. As controller 1 user, enable **Webform** content for microsite 1
9. As editor 1 on microsite 1, confirm I can create **Webform** content.
9. As controller 1 user, enable **news** content for microsite 1
10. As editor 1 on microsite 1, confirm I can create **news** content.
11. As controller 1 user, enable **events** content for microsite 1
12. As editor 1 on microsite 1, confirm I can create **events** content.
13. As controller 1 user, enable **Directories** content for microsite 1
14. As editor 1 on microsite 1, confirm I can create **Directories** content.
15. As controller 1 user, enable **blogs** content for microsite 1
16. As editor 1 on microsite 1, confirm I can create **blog** content.

ekes:
Requires https://github.com/localgovdrupal/localgov_geo/pull/118
Will fix #464 

To do:

 - [x] Should switch to use the user and group roles we define in the submodules. 
 - [ ] Add tests for **'Trusted editor'** role because users with this role are being given permission to create the content outside of the group.
 - [x] Why does Log Out not work? It does outside of tests.

Bonus:

 - Use logout to compare member and admin: Adding tests for global (non-group path) creation of content when it's included in the access per domain.


ekes:
Includes tests that will fail that are fixed by #470

finnlewis:
Need to test this locally, as 10.3 is not installing on our automated workflows.

ekes:
@stephen-cox What I noticed and was trying to fix with the commits there is that the project composer.json was requiring `@stable` so I made a branch that didn't. But when I put `feature/4.x-drupal-10.3` https://github.com/localgovdrupal/localgov_microsites_group/pull/469/commits/5ca117fa6f431ec090aa56629903265ee988649c it errors:

```
In VersionParser.php line 521:
                                                                               
  Could not parse version constraint feature/4.x-drupal-10.3: Invalid version  
   string "feature/4.x-drupal-10.3"
```

However, if I put in `dev-feature/4.x-drupal-10.3` or `feature/4.x-drupal-10.3-dev` it's going to append `-dev` to the string https://github.com/localgovdrupal/localgov_microsites_group/blob/6c870cd9a56b4dae4885e86a1f74086e757c4248/.github/workflows/localgov_microsites.yml#L51

So I guess... well it would be nicest if it did the appending automatically but didn't version fail? But otherwise should the lines to add -dev go? I'm now not sure.

But I think it is the version constraint on drupal core in the project composer.json that was causing the issue for not running the tests.

stephen-cox:
@ekes @finnlewis I have fixed testing against Drupal 10.3 with: https://github.com/localgovdrupal/localgov_microsites_group/pull/482

Do we want to merge this PR? Otherwise you can merge the `fix/test-10.3` branch into your feature branches.

finnlewis (approved):

finnlewis:
@stephen-cox this is the one we want to start with.

stephen-cox:
I've fixed the coding standards errors and there's another issue for the deprecated code issues #483 

The tests look correct to me and are failing because they pick up genuine issues. Are we just adding tests here or are we trying to fix stuff?

There are a load of PRs just adding tests and they are now all mixed up as well, which is quite confusing.

If we're just adding tests can we create one PR for them and close all the others? We could then detail all the issues found so we can start fixing them.

ekes:
Failing tests should be fixed in 470

> Includes tests that will fail that are fixed by https://github.com/localgovdrupal/localgov_microsites_group/pull/470

finnlewis:
> The tests look correct to me and are failing because they pick up genuine issues. Are we just adding tests here or are we trying to fix stuff?

Thanks @stephen-cox ! The intention was for those tests to fail here, so we can confirm #470 fixes them.

I've merged your commits from here to https://github.com/localgovdrupal/localgov_microsites_group/pull/470 / https://github.com/localgovdrupal/localgov_microsites_group/tree/feature/464-access-disabled-modules--plus

Let's close this one in favour of #470 

We agree that the tests are testing the right thing, which did break, and hopefully will not break in the future.

PR #478, opened 2024-06-01
Add blogs tests (and remove permissions not available from module).

ekes:


stephen-cox (approved):
This new test passes and looks fine to me.

PR #480, opened 2024-06-05
Check entities being inserted add them to group where appropriate.

ekes:
- If not being added in a group relationship path.
 - If being added on a group domain.
 - If is a content type that can be added to the group.

Fixes issues such as:
 - User has access to and has managed to navigated to /node/add or /media/add,
 - That media is uploaded in the modal and not immediately added to the group, thus sometimes lost.


finnlewis:
I'll test this now!

finnlewis:
Yes! This works for me @ekes !

Without this branch, I tested creating media on site1 at /media/add/image. The media was also visible on site 2.

With this branch, created media on site 1 and /media/add/image the media was only available in site 1.

I tested with nodes too,but they did not seem to exhibit the problem in the first place. Nodes created on site 1 and /node/add/page were only visible on site 1.

Happy to merge this!

finnlewis (approved):

PR #482, opened 2024-06-06
Run tests against Drupal 10.3

stephen-cox:


finnlewis:
Nice one @stephen-cox 

I'll create a separate issue for the deprecations, then let's merge this one shall we?

finnlewis (approved):
Captured the deprecations here: https://github.com/localgovdrupal/localgov_microsites_group/issues/483

finnlewis:
Noting that we are targeting Drupal core 10.3, as our tests do not work on 10.2.

PR #485, opened 2024-07-12
Fix: Inherited trait property declaration

Adnan-cds:
Calls to Drupal::service('domain_group_resolver') are causing WSOD :(  Error message says `PHP Fatal error:  Drupal\localgov_microsites_group\DomainGroupResolver and Drupal\group_context_domain\GroupFromDomainContextTrait define the same property ($domainNegotiator) in the composition of Drupal\localgov_microsites_group\DomainGroupResolver. However, the definition differs and is considered incompatible. Class was composed in /path/to/drupal/web/modules/contrib/localgov_microsites_group/src/DomainGroupResolver.php on line 24`

This is due to [recent changes](https://www.php.net/trait#language.oop5.traits.properties.example) to Drupal\group_context_domain\GroupFromDomainContextTrait

### DIFF
--- a/src/DomainGroupResolver.php
+++ b/src/DomainGroupResolver.php
@@ -29,20 +29,6 @@ class DomainGroupResolver implements DomainGroupResolverInterface {
   use GroupFromDomainContextTrait;
 
   /**
-   * The entity type manager.
-   *
-   * @var \Drupal\Core\Entity\EntityTypeManagerInterface
-   */
-  protected $entityTypeManager;
-
-  /**
-   * The domain negotiator service.
-   *
-   * @var \Drupal\Domain\DomainNegotiatorInterface
-   */
-  protected $domainNegotiator;
-
-  /**
    * DomainGroupHelper constructor.
    *
    * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager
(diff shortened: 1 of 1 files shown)


Adnan-cds:
Code [deprecation errors](https://github.com/localgovdrupal/localgov_microsites_group/actions/runs/9911371056/job/27384033842?pr=485#step:4:15) are unrelated.  Wondering if I should fix them.

src/DomainGroupResolver.php
```diff
@@ -54,7 +40,7 @@ class DomainGroupResolver implements DomainGroupResolverInterface {
    * @param \Drupal\Core\Entity\EntityRepositoryInterface $entity_repository
    *   Current entity repository interface.
    */
-  public function __construct(EntityTypeManagerInterface $entity_type_manager, DomainNegotiatorInterface $domain_negotiator, RouteMatchInterface $current_route_match, EntityRepositoryInterface $entity_repository) {
+  public function __construct(protected EntityTypeManagerInterface $entity_type_manager, protected DomainNegotiatorInterface $domain_negotiator, protected RouteMatchInterface $current_route_match, protected EntityRepositoryInterface $entity_repository) {
```
ekes: If we're declaring theme here (`protected EntityTypeManager Interface ...` etc) we should switch to `protected EntityTypeManagerInterface $entityTypeManager` however if the Trait is declaring them I suspect we need to just remove our declaration of them, and set them, not declaring `$entity_type_manager` as a $this variable.

Adnan-cds: > not declaring $entity_type_manager as a $this variable.

Right.

Adnan-cds: > I suspect we need to just remove our declaration of them

We have two options here.  We can keep the declarations or remove them.  I am not sure of the merit of either.  You seem to prefer to remove the declaration.s  So I will just go with that.

ekes:
So running through for clarity:

 * $this->entityTypeManager is defined by GroupRouteContextTrait (and DomainFromGroupTrait both of which are in sync for how the property is defined - it was DomainGroupResolver defining it differently that caused the error in the first instance I think)
 * $this->domainNegotiator is defined by GroupFromDomainContextTrait
 * $this->currentRouteMatch is defined by GroupRouteContextTrait
 * $this->entityRepository is defined by GroupFromDomainContextTrait.

As this is deprecated code, and we can make sure the Traits elsewhere sync up their definitions if need be, leaving them out completely seems a reasonable option here.

ekes (approved):
See comment on issue.
Remaining deprecated code check is from D12.

Adnan-cds:
Thanks.  Merging...

PR #488, opened 2024-07-17
Rework attempt for rendering and listing of themes to remove deprecat…

finnlewis:
…ed methods.

### DIFF
--- a/src/Plugin/DomainGroupSettings/ThemeSettings.php
+++ b/src/Plugin/DomainGroupSettings/ThemeSettings.php
@@ -34,7 +34,7 @@ class ThemeSettings extends DomainGroupSettingsBase implements ContainerFactoryP
   /**
    * {@inheritdoc}
    */
-  public function __construct(array $configuration, $plugin_id, $plugin_definition, protected ConfigFactoryInterface $config_factory, protected EntityTypeManagerInterface $entity_type_manager, protected LanguageManagerInterface $language_manager, protected ThemeHandlerInterface $theme_handler, protected ThemeExtensionList $extension_list_theme) {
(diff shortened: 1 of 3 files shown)


src/Plugin/DomainGroupSettings/ThemeSettings.php
```diff
@@ -31,36 +31,11 @@ class ThemeSettings extends DomainGroupSettingsBase implements ContainerFactoryP
...
-   *
-   * @var \Drupal\Core\Extension\ThemeHandlerInterface
-   */
-  protected $themeHandler;
-
   /**
    * {@inheritdoc}
    */
-  public function __construct(array $configuration, $plugin_id, $plugin_definition, ConfigFactoryInterface $config_factory, EntityTypeManagerInterface $entity_type_manager, LanguageManagerInterface $language_manager, ThemeHandlerInterface $theme_handler) {
+  public function __construct(array $configuration, $plugin_id, $plugin_definition, protected ConfigFactoryInterface $configFactory, protected EntityTypeManagerInterface $entityTypeManager, protected LanguageManagerInterface $languageManager, protected ThemeHandlerInterface $themeHandler, protected ThemeExtensionList $extensionListTheme) {
```
ekes, line 37: Looks good. It's only if it screams about the redeclaration of entityTypeManager which is declared in one, or more, of the Traits at a guess. But it'll scream pretty early on.

finnlewis: Thanks @ekes  - seems to be happy locally, so I'll merge and we can see how it plays with the tests.

PR #489, opened 2024-07-19
202407 media search name

anthonylindsay:
added group node admin view override 
- adds a 'search by name' filter

PR #486, opened 2024-07-16
Adds reset button and author column for group node view

anthonylindsay:
config change for new installs for admin view of nodes for a microsite.

PR #479, opened 2024-06-02
Feature/473 webform submission test

ekes:
NB it relies on #469 as it has the login out trait.

PR #491, opened 2024-07-30
4.0.0-beta5

ekes:


ekes:
Looks like some of the last commits might have been squashed and merged into the release branch (maybe as they weren't squashed and merged into the 4.x branch).

Actual differences here are:
 * #479 webform test
 * #486 group node view author column
 * #489 media search view

Issue #492, opened 2024-08-06
Errors on view goup media schema in microsites group tests


ekes:
Drupal\Tests\localgov_microsites_group\Kernel\ContentTypeHelperTest::testToggleModule
Exception: Exception when installing config for module group, message was: Schema errors for views.view.group_media with the following errors: views.view.group_media:display.default.display_options.filters.name.expose.autocomplete_filter missing schema, views.view.group_media:display.default.display_options.filters.name.expose.autocomplete_min_chars missing schema, views.view.group_media:display.default.display_options.filters.name.expose.autocomplete_items missing schema, views.view.group_media:display.default.display_options.filters.name.expose.autocomplete_field missing schema, views.view.group_media:display.default.display_options.filters.name.expose.autocomplete_raw_suggestion missing schema, views.view.group_media:display.default.display_options.filters.name.expose.autocomplete_raw_dropdown missing schema, views.view.group_media:display.default.display_options.filters.name.expose.autocomplete_dependent missing schema, views.view.group_media:display.default.display_options.filters.name.expose.autocomplete_contextual missing schema, views.view.group_media:display.default.display_options.filters.name.expose.autocomplete_autosubmit missing schema

/var/www/html/web/core/tests/Drupal/KernelTests/KernelTestBase.php:770
/var/www/html/web/modules/contrib/group/tests/src/Kernel/GroupKernelTestBase.php:36
/var/www/html/web/modules/contrib/localgov_microsites_group/tests/src/Kernel/ContentTypeHelperTest.php:64
/var/www/html/vendor/phpunit/phpunit/src/Framework/TestResult.php:729

ekes:
Is it related to https://github.com/localgovdrupal/localgov_microsites_group/pull/489

PR #493, opened 2024-08-09
Fix schema error #492

stephen-cox:
## What does this change?

Removes invalid schema from group media view.

Adds GitHub shared workflow to fix test runs.

## How to test

See the tests pass. Check that the site media listing still works after an install.


ekes (approved):
Tests pass!

Plus it removes probably additional dependency config which is good.
Assume the test file updates are good.

finnlewis (approved):

ekes:
> Assume the test file updates are good.

No longer assume. I see they are.

PR #495, opened 2024-08-19
Patch group module to support previewing content.

finnlewis:
## What does this change?

Add's a patch : 

https://www.drupal.org/files/issues/2023-09-25/group2-2853001-26.patch

From: 

http://drupal.org/node/2853001

## How to test

For now, I'm just testing this patch to see if it helps with the preview_link and responsive_preview functionality.


stephen-cox:
@finnlewis I'm not sure this patch is needed. Content preview works fine on group content and Preview Link also looks to work once Trusted Editors are given then 'Generate preview links' permission. Or am I missing something?

finnlewis:
@stephen-cox I think you are right!

PR #497, opened 2024-09-18
Add patch from https://www.drupal.org/project/group/issues/2853001 to…

finnlewis:
This an issue, again!

Just like mentioned here: https://www.drupal.org/project/responsive_preview/issues/3302861

Which takes us back to https://www.drupal.org/project/group/issues/2853001

The latest patch on there seems to suppress the problem.

Steps to reporoduce:

install localgov_microsites 4.0.0 :)
create a microsite and assign an admin user to the microsite
log into microsite as admin user
create a newsroom
See

Error message

> Warning: Undefined array key "preview" in Drupal\responsive_preview\ResponsivePreview->formAlter() (line 221 of modules/contrib/responsive_preview/src/ResponsivePreview.php).
> 
> Warning: Trying to access array offset on value of type null in Drupal\responsive_preview\ResponsivePreview->formAlter() (line 221 of modules/contrib/responsive_preview/src/ResponsivePreview.php).
> 

Solution: apply this patch https://www.drupal.org/files/issues/2023-09-25/group2-2853001-26.patch

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,6 +43,9 @@
             },
             "drupal/ginvite": {
                 "View schema fix #3324625": "https://www.drupal.org/files/issues/2022-11-30/3324625-2-schema-fix-for-4.0.x.patch"
+            },
+            "drupal/group": {
+                "Fix Warning: Undefined array key 'preview' in formAlter(). https://www.drupal.org/project/group/issues/2853001": "https://www.drupal.org/files/issues/2023-09-25/group2-2853001-26.patch"
             }
         }
     }


finnlewis:
@ekes or @stephen-cox can you help me work out why this test is failing? 

When I test it manually, it behaves as expected. 

Logging into the control site forwards the user to /admin

Logging into a microsite forwards the user to /group/1

So why would the test be behaving differently?

https://github.com/localgovdrupal/localgov_microsites_group/actions/runs/10921854523/job/30315030208?pr=497#step:5:16

```
1) Drupal\Tests\localgov_microsites_group\Functional\LoginTest::testLoginRedirect
Behat\Mink\Exception\ExpectationException: Current page is "/user/2", but "/admin" expected.

/var/www/html/web/core/tests/Drupal/Tests/WebAssert.php:591
/var/www/html/web/core/tests/Drupal/Tests/WebAssert.php:779
/var/www/html/web/modules/contrib/localgov_microsites_group/tests/src/Functional/LoginTest.php:80
/var/www/html/vendor/phpunit/phpunit/src/Framework/TestResult.php:729
```

ekes:
> So why would the test be behaving differently?

If the same code passes locally, and fails here. I think something about the way the test is run meaning that the domain negotiator does not think there is an active domain:-

https://github.com/localgovdrupal/localgov_microsites_group/blob/21cdf8be54f0b0efe104c3348eecb23c6df01be9/localgov_microsites_group.module#L287

that's the reason it wouldn't redirect to either admin or the group page.

or it getting cached differently
https://git.drupalcode.org/project/domain/-/blob/2.0.x/domain/src/DomainNegotiator.php?ref_type=heads#L146
but the negotiation is
https://git.drupalcode.org/project/domain/-/blob/2.0.x/domain/src/DomainNegotiator.php?ref_type=heads#L164

I guess in group site tests we set the site, is it not defaulting to having the _domain_ of the control site at other times? How is that set when running the test, environment variables?

ekes:
Guess this is where it's setting the domain https://github.com/localgovdrupal/localgov_shared_workflows/blob/5db7a020b4e724a81ac9a4e2750342facba53ee7/.github/workflows/test-microsites.yml#L223

ekes:
So locally it installs with
```
uuid: 6ddda674-911f-4efb-ad0c-2da28bca56c2
langcode: en
status: true
dependencies: {  }
id: localgov_microsites_ddev_site
domain_id: 2467579
hostname: localgov-microsites.ddev.site
name: 'Drush Site-Install'
scheme: variable
weight: -1
is_default: true
```

And during a test the ID is: `example_com` but the hostname: is updated to web or whatever I'm asking for; and it's the default.

What is it in the github test, and why isn't it returned if it's the default anyway. What's creating and updating this.

finnlewis:
I think we should merge and release this, and follow up on the odd failing tests on https://github.com/localgovdrupal/localgov_microsites_group/issues/503

ekes (approved):
Just read the issue. As I understand it this patch fixes the issue for us for now, but what will finally get merged might be a different approach. But I don't think changing from one to the other is going to be an issue, and people on the issue seem to be happy enough with this approach now too.

PR #506, opened 2024-10-04
Hide front and error page settings from microsite controllers 

### LINKED ISSUE #486: Controller cannot save microsite site-stettings
As a controller logged into the control site
I want to save the site-settings form for a microsite
So that I can configure the microsite from the control site

But we can't!

Steps to replicate

1. install microsites 4.0.0
2. create a user with the 'controller' role 
3. log into the control site as the controller user
4. create a microsite
5. edit the microsite
6. Edit the site settings https://localgov-microsites.ddev.site/group/1/domain-settings
7. Save
8. See "Error message 1 error has been found:    [Site front page](https://localgov-microsites.ddev.site/group/1/domain-settings#edit-site-frontpage)

image

This seems to happen because the controller user does not have access to the microsite content on the control site, so Drupal is not happy setting the front page.

Maybe we should just hide the domain-settings from the control site?

Or remove the 'site front page' and 403 and 404 page fields from the form on the control site.

@ekes @stephen-cox any thoughts on the best logic here?

stephen-cox:
Fixes localgovdrupal/localgov_microsites#486

## What does this change?

Removes the Site front page and error pages fields from the site settings form for a microsite for microsite controllers as they don't have permission to access these.

This checks whether the user has the Microsites Controller role, but I realise it maybe better to check for the relevant group content permissions. Let's discuss.

## How to test

Check the fields are there for site admins but not for controllers.

### Site admin

image

### Microsite controller

image

### DIFF
--- a/src/Form/DomainGroupSettingsForm.php
+++ b/src/Form/DomainGroupSettingsForm.php
@@ -92,7 +92,7 @@ class DomainGroupSettingsForm extends FormBase {
     ];
 
     // Hide configuration options that site controllers don't have access to.
-    if (in_array('microsites_controller', $this->currentUser->getRoles())) {
+    if (!$group->hasPermission('access group_node overview', $this->currentUser)) {
       $form['domain_group_site_settings']['error_page']['#access'] = FALSE;
       $form['domain_group_site_settings']['site_frontpage']['#access'] = FALSE;
     }
--- a/src/Form/DomainGroupSettingsForm.php
+++ b/src/Form/DomainGroupSettingsForm.php
@@ -16,6 +16,13 @@ use Symfony\Component\DependencyInjection\ContainerInterface;
 class DomainGroupSettingsForm extends FormBase {
 
   /**
+  * Current user account.
+  *
+  * @var \Drupal\Core\Session\AccountInterface
+  */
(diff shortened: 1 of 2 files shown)


finnlewis:
@stephen-cox mentioned we may want to change this to check for permissions to manage content in the group, rather than just the role.

ekes (approved):

finnlewis:
On an existing install, testing this branch, going to that page and saving gives me: 

```
 The path '/node/2' is either invalid or you do not have access to it. 
```


image

I wonder do we also need to hook into something else to prevent it trying to save the original value?

finnlewis:
Also on a fresh install, create a controller user. create a single microsite, edit and save the 'Site settings' tab.

```
The path '/welcome-your-new-site' is either invalid or you do not have access to it. 
```

image

stephen-cox:
@finnlewis Should be fixed now - I have completely removed the fields and it works for me, although I'm sure it worked for me previously.

ekes:
Tests now pass, do we need @finnlewis to see if the issue can be recreated/has been fixed?

finnlewis:
Works for me now, thanks @stephen-cox :)

PR #510, opened 2024-10-18
Limit the accessible domain path settings to the active domain 

### LINKED ISSUE #509: URL aliases available for all microsites on a given piece of content
For example: 
image

@stephen-cox recalls writing some code to fix this... maybe we need to fond that and update it?

stephen-cox:
## What does this change?

Hides the path settings for all the domains apart from the active one.

## How to test

When editing a page on a site you should only be able to see path settings for the site you're on.

When creating a new node the generate automatic URL checkbox should be checked.

Fixes #509

### DIFF
--- a/localgov_microsites_group.module
+++ b/localgov_microsites_group.module
@@ -9,6 +9,7 @@ use Drupal\Component\Serialization\Yaml;
 use Drupal\Core\Cache\Cache;
 use Drupal\Core\Cache\RefinableCacheableDependencyInterface;
 use Drupal\Core\Entity\Display\EntityViewDisplayInterface;
+use Drupal\Core\Entity\EntityFormInterface;
 use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Form\FormStateInterface;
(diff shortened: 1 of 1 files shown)


stephen-cox:
I have run out of time today to add tests to this. As I thought I had fixed this, but can't see any sign of that, I think it would be worth adding a test to ensure it keeps working.

finnlewis:
Tested locally and it works a treat, thanks @stephen-cox !

Do we want to write some tests before merging?

millnut:
Looks like the failed tests here and other PRs mentioned today are unrelated as we get them on the main project from this morning's run https://github.com/localgovdrupal/localgov_microsites_project/actions/runs/11454313013/job/31868361713 looks like something needs updating in the `group_sites` module.

Looks to have started on this run (19th Oct) https://github.com/localgovdrupal/localgov_microsites_project/actions/runs/11415200204

finnlewis:
Follow up issue to create tests https://github.com/localgovdrupal/localgov_microsites_group/issues/516

PR #514, opened 2024-10-24
Ensure deprecated variationcache module is included and then uninstall it

### LINKED ISSUE #513: Currently using Missing or invalid modules: variationcache
After upgrading to Group 3.3.0 I see the following error:

The following modules are marked as installed in the core.extension configuration, but they are missing: 

 * variationcache

Group dropped the requirement for this with https://git.drupalcode.org/project/group/-/commit/4349a9f1c0779db024b5657145f139bb5db213bf and the module is deprecated as the functionality was included in Drupal 10.2: https://www.drupal.org/project/variationcache

I suggest we add a requirement for variation cache to our composer.json, but include an update hook to uninstall the module. We can then drop the requirement for this module with a later release.

Unless anyone has a better way to deal with this?

stephen-cox:
Fixes #513

## What does this change?

The deprecated variationcache module is removed as a dependency from Group. As it's no longer required ensure it exists and uninstall it.

## How to test

Check variationcache module is uninstalled and the site still functions.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,6 +16,7 @@
         "drupal/override_node_options": "^2.6",
         "drupal/replicate": "^1.0",
         "drupal/role_delegation": "^1.2",
+        "drupal/variationcache": "^1.5",
         "localgovdrupal/group_webform": "^1.0.0-beta2",
         "localgovdrupal/localgov_core": "^2.12",
         "localgovdrupal/localgov_forms": "^1.0.0-beta2",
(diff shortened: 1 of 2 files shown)


stephen-cox:
Test failures look to be the same as the ones effecting the 4.x branch. I'll try and fine some time to look at the failures later today anyway.

finnlewis (approved):
Sweet!

ekes:
We should have bumped the dependency on group to 3.3.x here. Attempting to uninstall variationcache with 3.2.x doesn't work. Follow-up on its way.

PR #522, opened 2024-11-12
Control site access policy, retain permissions from role. Fixes #521

### LINKED ISSUE #521: Controller denied access to groups created by User 1
Steps to reproduce:

1. Install Microsites
2. Login as Drupal User 1
3. Create a Microsite
4. Create a Controller user
5. Login as Controller
6. No Microsites visible

I know we say not to do this, but it used to work and doesn't any longer. I think it's due to a recent change in Group Sites.

Changing the Group Sites access policy when there is no group (/admin/group/sites/settings) from our custom handler to 'Do nothing' fixes the issue.

image

ekes:
Previous control site permissions were too restrictive, removing all permissions not available to the individual. This allows back the permissions we want on the control site if they are also granted by role.


stephen-cox (approved):
Fixes issue #521 for me.

Issue #526, opened 2024-11-22
Require Group 3.3.x


ekes:
We should have bumped the dependency on group to 3.3.x here. Attempting to uninstall variationcache with 3.2.x doesn't work. Follow-up on its way.

_Originally posted by @ekes in https://github.com/localgovdrupal/localgov_microsites_group/issues/514#issuecomment-2494084335_

stephen-cox:
Fixed with #527

PR #527, opened 2024-11-22
[#526] fix: Require Group 3.3.x when removing variationcache.

ekes:
Shouldn't remove variationcache with 3.2.x installed still.

3.2 depends on flexible_permissions 1.x which is still a full module depending on variationcache (even if it's just the 1.5 stub version).
3.3 depends on flexible_permisisons 2.x which is a stub module removing the dependency on variationcache.


finnlewis (commented):
Makes sense, thanks @ekes !

finnlewis (approved):

PR #528, opened 2024-11-22
Add field_domain_access value to domain_path forms

stephen-cox:
The Domain Path module does a lot of checks for domain access configuration, but we use the Group module for access checks. Adding the domain access field to the form ensures the Domain Path checks pass, but is a bit of a hack as it would be better to fix whatever is going wrong in the Domain Path module.

## What does this change?

Adds a field_domain_access value field to the forms with domain_path field to ensure that logic that does domain access checks pass.

## How to test

Saving paths on microsite nodes should now work.

## Have we considered potential risks?

This is a bit if a hack. It would be better to work out what's going wrong in the Domain Path module and patch it.

This most definitely needs tests. These would be necessary even if we patched the Domain Path module.

### DIFF
--- a/localgov_microsites_group.module
+++ b/localgov_microsites_group.module
@@ -243,9 +243,25 @@ function localgov_microsites_group_form_alter(&$form, FormStateInterface $form_s
       if ($id !== $active_domain->id()) {
         $form['domain_path'][$id]['#access'] = FALSE;
       }
+      else {
+
+        // The Domain Path module does a lot of checks for domain access
+        // configuration, but we use the Group module for access checks. Adding
+        // the domain access field to the form ensures the Domain Path checks
+        // pass, but is a bit of a hack as it would be better to fix whatever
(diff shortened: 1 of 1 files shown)


ekes:
If this works maybe we should get it in. I was chatting to webflo at DrupalCamp Berlin, he's taken on some maintaining Domain module, and was well up for people working on domain_path (particularly, or even specifically, the pathauto integration)... so getting a fix upstream applied upstream if we can make one would seem like a good idea.

finnlewis:
I tested this, it works for me!

finnlewis:
@stephen-cox wants to add some tests for this.
I think we should merge this and release as a fix for the people that need it.

Then: 

1. add an issue to write tests to cover this. https://github.com/localgovdrupal/localgov_microsites_group/issues/530 
2. add an issue to refactor this by perhaps patching domain_path module to cope with our situation. https://github.com/localgovdrupal/localgov_microsites_group/issues/531

PR #533, opened 2024-11-26
Added tests for setting domain path aliases

### LINKED ISSUE #530: Add tests for manual path aliases on microsites
Coming from https://github.com/localgovdrupal/localgov_microsites_group/pull/528 

@stephen-cox keen to make sure we have tests that confirm : 

 - only the active site alias is displayed
 - manually setting a path alias works
 - editing a path alias works
 - path auto also works.

@stephen-cox has work in progress on this.

stephen-cox:
Fixes #530 

## What does this change?

Adds tests for setting and editing domain path aliases on microsite content.

## How to test

Tests pass

## How can we measure success?

Domain path stuff doesn't break again without us being notified.

### DIFF
--- a/tests/src/Functional/DomainPathAliasTest.php
+++ b/tests/src/Functional/DomainPathAliasTest.php
@@ -1,0 +1,155 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Drupal\Tests\localgov_microsites_group\Functional;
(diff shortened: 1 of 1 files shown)


finnlewis (approved):
Delicious, thanks @stephen-cox !

PR #545, opened 2025-07-25
Disable per-domain access control while indexing.

ekes:
This populates the index with all items and their referenced items. The items are filtered on output.

See https://github.com/localgovdrupal/localgov_microsites/issues/524

**Requires** https://github.com/localgovdrupal/localgov_directories/pull/442 to work.

### DIFF
--- a/localgov_microsites_group.services.yml
+++ b/localgov_microsites_group.services.yml
@@ -73,3 +73,9 @@ services:
     tags:
       - { name: 'context_provider' }
     deprecated: 'The "%service_id%" service is deprecated. You should use the group sites context directly instead.'
+
+  localgov_microsites_group.event_subscriber:
+    class: Drupal\localgov_microsites_group\EventSubscriber\SearchApiSubscriber
+    arguments: ['@group_sites.admin_mode']
+    tags:
+      - { name: event_subscriber }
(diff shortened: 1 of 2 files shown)


nccchris (commented):
The search now finds the examples I posted and newly created content: i.e. it's working as expected.

finnlewis:
Working, but would like to understand why it works. @ekes  can you explain it to me again?

ekes:
As it starts indexing https://github.com/localgovdrupal/localgov_microsites_group/pull/545/files#diff-1ac91beafa0a59e0f8d8af8314c37c695d638d9faaff47870e691ef497c42693R33 it puts group sites into admin mode - it's not enforcing access per domain - because we index all the sites at the same time from one (or no) domain. Then when it finishes https://github.com/localgovdrupal/localgov_microsites_group/pull/545/files#diff-1ac91beafa0a59e0f8d8af8314c37c695d638d9faaff47870e691ef497c42693R40 it unsets admin mode.

It needs the other patch because this hook https://github.com/localgovdrupal/localgov_directories/pull/442/files#diff-38d0ee7d906a81c80b91deeb640c8c484e11bb16dc28f38a841274c08212b7e1L295 in which we load the fields is run earlier than the event subscribed to above. It then has the per-site domain access control and doesn't load all the items. Pushing it also into a processor https://github.com/localgovdrupal/localgov_directories/pull/442/files#diff-06c1c67ce91f831b4328ad4694599959f80ff7abc4b809112e02d3f6af095a96R24 means it happens after our subscribed start index event has happened.

finnlewis (approved):

PR #556, opened 2025-09-05
adds publications to microsites

### LINKED ISSUE #555: Add Publications to microsites

markconroy:
Closes #555  

---

Not completed yet. This add the config for publications, but _does not_ have the "Book Outline" field in the node create/edit sidebar.

---

## What does this change?

Configures publications module for microsites.

## How to test

- Install the new localgov_microsites_group submodule
- Create a microsite
- See that you can add publications overview and publications pages to the microsite

## How can we measure success?

People no longer ask us to add Publications to Microsites.

---
Thanks to [Big Blue Door](https://www.bigbluedoor.net/) for sponsoring my time to work on this.


finnlewis:
@markconroy we've got a typo here I think, picked up by the coding standards checks:
 
https://github.com/localgovdrupal/localgov_microsites_group/actions/runs/17489549389/job/49676141135?pr=556#step:4:24

finnlewis:
@markconroy mentions that the book outline is not showing up in the meta data sidebar.
Also we have conflicts to resolve

finnlewis:
Testing locally and yes, the publication outline is a problem.

On the control site, I can see this in the sidebar: 

image

On the microsite, I cannot.

So something group permission related I guess.

I will investigate further.

Also need to place the blocks for publication header and publication navigation.

markconroy:
> Also need to place the blocks for publication header and publication navigation.

Blocks are now placed.

finnlewis:
Worth looking at: 

 - https://www.drupal.org/project/groupbook
 - https://www.drupal.org/project/book
  
Might be worth upgrading to recent version of book.

finnlewis:
@rupertj have you successfully used localgov_publications in microsites?

rupertj:
I've yet to try it, TBH.

I don't think placing the blocks in this module is the right thing to do. Publications already tries to place the blocks using the block installer. If that's not happening on microsites, it'd be worth figuring out why. We could also place it the blocks in all the active themes for all the microsites.

AWearring:
I've just tested this on my local build and I get the Publication block
image

AWearring:
I can also confirm that child pages can be added to the parent publication and these pages aren't available to another domain on the same build

finnlewis:
> I can also confirm that child pages can be added to the parent publication and these pages aren't available to another domain on the same build

What am I missing then? 

I still do not see the book outline when adding a publication page on a microsite.

finnlewis:
Aha - it looks like these permissions are missing: 
image

finnlewis:
Might just need to add permission's for the book module?

markconroy:
@finnlewis Wanna give this another spin?

rupertj:
The issue stalled a bit, but the plan is to stop publications granting "administer book outlines". See https://github.com/localgovdrupal/localgov_publications/issues/83.

rupertj:
It also looks like there's some unrelated permissions in there. I don't think the ones that mention "group_term:localgov_topic" are anything to do with publications.

markconroy:
Topic permissions have been removed now.

rupertj:
Then for the blocks, I'm pretty sure that if you add localgov_microsites_base into this array here https://github.com/localgovdrupal/localgov_core/blob/2.x/src/Service/DefaultBlockInstaller.php#L63 you can get rid of the duplicated block config.

That won't work for microsites that have their own theme and block configuration, but then the way this PR currently does it won't either. We could make that happen in localgov_core though.

rupertj:
I'm also thinking that we could get rid of the whole concept of these submodules that integrate another module with microsites pretty easily by moving the microsites permissions hook to the module being supported and making the group relationship config optional, with a dependency on localgov_microsites_group, but that's probably out of scope for this PR. (Unless we want publications to be the guinea pig for that...)

markconroy:
Thanks @rupertj you were right about adding `localgov_micriotes_base` to the array in `localgov_core`.

Here's the related PR for LocalGov Core: https://github.com/localgovdrupal/localgov_core/pull/324

finnlewis:
Testing this again to see where we're at.

Any recent updates @markconroy @rupertj @tonypaulbarker ?

finnlewis:
This does seem to work well now :)

The publications (book) fields show in the sidebar for a trusted editor user on a microsite.

1. Fresh install of localgov_microsites
2. enable localgov_microsites_publications
3. create controller user
4. log in as controller user
5. create two microsites and a site admin with trusted editor role
6. log into microsite 1 as the site admin
7. create publication pages
8. confirm I can create new publication outline 
9. confirm I can add more pages to the publication
10. confirm these are not available on microsite 2

@rupertj @markconroy I do see "Breadcrumbs" listed in the "On this page" navigation. Is that a known issue? 

image

rupertj:
It's new to me. Is there a hidden heading with text "Breadcrumbs" in the source?

rupertj:
It's doing that because localgov_microsites_base has the breadcrumbs block hardcoded into the node-full template.

finnlewis:
On the breadcrumb issue, sounds like this is related so we can pick it up there and resolve  https://github.com/localgovdrupal/localgov_microsites_base/issues/338

finnlewis:
Just chatting with @tonypaulbarker and we're happy to merge this and fix the breadcrumbs issue on https://github.com/localgovdrupal/localgov_microsites_base/issues/338

Thanks @markconroy @rupertj and @AWearring :)

PR #566, opened 2025-10-23
4.3.0 release

finnlewis:
* adds publications to microsites


finnlewis:
Re-running the tests on this. Just one random cache check failing in PHP 8.2. Odd that it would fail in 8.2 but pass in 8.1 and 8.3 🤷

